PR #1, opened 2015-09-18
For those that don't have JULIAHOME defined, autogenerate it!

staticfloat:
This doesn't check to make sure people are in a source checkout, I was too lazy.


timholy:
Great, thanks!

PR #8, opened 2016-04-01
Handle # characters in function-types (julia 0.5). Fixes #7

### LINKED ISSUE #7: Not Very useful on 0.5
0.5 signatures contain a lot of calls with `#` in them, which SnoopCompile ignores.

timholy:
Replaces `Base.#any` with `typeof(Base.any)`.

### DIFF
--- a/src/SnoopCompile.jl
+++ b/src/SnoopCompile.jl
@@ -136,7 +136,17 @@ end
 function parse_call(c; subst = Dict(), blacklist=UTF8String[])
     fpath = [""]
     args = ""
-    contains(c, "#") && return false, c, fpath, args    # skip gensyms
+    hash_handled = match(r"#[0-9]", c) == nothing  # gensyms and anonymous functions on 0.5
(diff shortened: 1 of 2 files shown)


timholy:
CC @Keno.

timholy:
Let me know if this works.

Keno:
I think you'll still run into trouble with anonymous functions which are `#42` or so.

timholy:
Now you're being greedy. I doubt there's any way around that problem; even if you extracted the AST associated with that function, constructed a new anonymous function that does the same thing, and issued a precompile directive, it wouldn't help because the name in a new session wouldn't match.

Keno:
I'm not saying it should work for anonymous functions, I'm saying this approach will produce invalid output on them.

timholy:
Ah, got it. Thanks for pointing that out, and already fixed (hopefully). Of course one can edit the resulting file manually, but you're right that there's no good reason not to exclude those automatically.

Feel free to give this a whirl and let me know if you have problems.

PR #14, opened 2017-11-03
revise for v0.6

vtjnash:
support is now in base, so snoop_on/off is not needed
also by careful parsing, we can avoid needing to use eval
and provide better support for non-leaf signatures and anonymous functions


timholy:
:100: 

I'm getting
```julia
ERROR: LoadError: failed parse of line: Tuple{typeof(Base.getindex), Tuple, Base.AbstractUnitRange{#s45} where #s45<:Real}
```
on the `images.jl` example. It's failing because of the `#`. The old version skipped gensyms, but we could presumably replace them with a random string?

vtjnash:
We could also just not throw that error when it encounters corrupt output from Base. Although I'm surprised that method definition got through: it's not particularly useful as a specType signature, so we shouldn't be trying to compile it.

examples/images.jl
```diff
@@ -4,32 +4,28 @@ using SnoopCompile
...
-# IMPORTANT: we must have the module(s) defined for the parcelation
-# step, otherwise we will get no precompiles for the Images module
-using Images
-
 data = SnoopCompile.read("/tmp/images_compiles.csv")
 
 # The Images tests are run inside a module ImagesTest, so all
 # the precompiles get credited to ImagesTest. Credit them to Images instead.
-subst = Dict("ImagesTests"=>"Images")
+subst = ["ImagesTests"=>"Images", ]
```
timholy: (Images no longer uses a module for its tests, but this is still useful as a demo.)

examples/images.jl
```diff
@@ -4,32 +4,28 @@ using SnoopCompile
...
 
 # Blacklist helps fix problems:
 # - MIME uses type-parameters with symbols like :image/png, which is
 #   not parseable
-blacklist = ["MIME"]
+blacklist = ["MIME", ]
```
timholy: The output format seems to have improved here (it's now `Expr(:call, :MIME, "image/svg+xml")`) so this is now parseable. Still might be useful as a demo?

src/SnoopCompile.jl
```diff
@@ -1,275 +1,217 @@
...
+        line = string(curly)
     end
-    pcstring = string("    precompile(", fname, ", ", args, ")")
-    true, pcstring, fpath, args
+    return true, line, topmod
 end
 
 """
-`pc, discards = parcel(calls; subst=Dict(), blacklist = UTF8String[])` assigns each compile statement to the module that owns the function. Perform string substitution via `subst=Dict("Module1"=>"Module2")`, and omit functions in particular modules with `blacklist=["Module3"]`.  On output, `pc["Module2"]` contains all the precompiles assigned to `Module2`, and `discards` contains functions that had arguments spanning modules and which therefore could not be assigned to a particular module.
+`pc, discards = parcel(calls; subst=[], blacklist=[])` assigns each compile statement to the module that owns the function. Perform string substitution via `subst=["Module1"=>"Module2"]`, and omit functions in particular modules with `blacklist=["Module3"]`. On output, `pc[:Module2]` contains all the precompiles assigned to `Module2`.
```
timholy: eliminate `discards` from docstring

src/SnoopCompile.jl
```diff
@@ -1,275 +1,217 @@
...
+    discards = String[]
     for c in calls
-        try
-            keep, pcstring, fpath, args = parse_call(c, blacklist=blacklist)
-        catch
-            println("error processing ", c)
-            rethrow()
-        end
+        local keep, pcstring, topmod
+        keep, pcstring, topmod = parse_call(c, blacklist=blacklist)
```
timholy: doesn't pass `subst`

timholy:
By the way, while I have you here, is there benefit to precompiling the following (examples from what the `images.jl` demo assigns to the `AxisArrays` package):
```julia
precompile(Tuple{typeof(AxisArrays.axisvalues), AxisArrays.Axis{:x, Base.StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}}}})
precompile(Tuple{typeof(AxisArrays.axisdim), Type{AxisArrays.AxisArray{ColorTypes.RGB{Float64}, 2, Array{ColorTypes.RGB{Float64}, 2}, Tuple{AxisArrays.Axis{:y, Base.OneTo{Int64}}, AxisArrays.Axis{:x, Base.OneTo{Int64}}}}}, Type{AxisArrays.Axis{:x, T} where T}})
```
The first involves a combination of `AxisArrays` + `Base`; the second a combination of `AxisArrays` + `ColorTypes` + `Base`. I recall an old conversation with you indicating that cross-module precompiles aren't cached (not sure whether that applies to `Base`). If one or both are not useful, should we strip them at this stage? Otherwise they just add to the build time without benefit.

vtjnash:
They're all useful for my current usage (`format_userimg`). They're perhaps less useful for building .ji files in the second example (where the expression would only be valid in a context where you have both ColorTypes and AxisArrays defined), but that won't affect compile times.

I've been using a macro similar to the following to filter those out:
```julia
macro precompile(expr)
    expr = esc(expr)
    return quote
        try; precompile($expr); catch e; println(STDERR, "could not precompile: ", expr); Base.show_error(STDERR, e); end
    end
end
```

KristofferC:
I guess the README could use an update as well?

timholy:
Thank you!

I'll tackle the README, thanks for pointing it out @KristofferC.

tknopp:
great to see progress here! Whats the situation for intermodule precompiles? Is there an issue where this is tracked / discussed?

timholy:
> They're perhaps less useful for building .ji files in the second example (where the expression would only be valid in a context where you have both ColorTypes and AxisArrays defined), but that won't affect compile times.

Suppose we choose a different situation where module A depends (via `using`) on module B. I'm unclear on whether you get a benefit from
```julia
precompile(Tuple{typeof(someAfunction), SomeBType})
```
My impression based on past conversations was "no, this isn't cached in the `.ji` file."

PR #18, opened 2018-08-26
v1.0 compatibility changes

dalum:
I haven't addressed the testing, and the package probably needs a `Project.toml` file, etc. too, but the changes here let me precompile my package successfully using `@snoop1` and include the resulting produced files with a noticeable effect on 1.0.

I ran into a number of issues related to package availability, though, trying to use `@snoop`, since the new Pkg has those projects, and people will probably typically be snooping from within a project.  So I hacked in the `using Pkg; Pkg.activate(".")`, which is definitely not the most stable solution.

Another thing is the `ccall(:jl_generating_output, Cint, ()) == 1 || return nothing` which I see from git blame has been there since the initial commit. What is the check for? On my machine, `ccall(:jl_generating_output, Cint, ())` always returns `0`, so I changed it to that, in order for the `_precompile_()` call to do any precompilation.

(Edit: just noticed I named my branch 0.1 instead of 1.0. :woman_facepalming:)

### DIFF
--- a/src/SnoopCompile.jl
+++ b/src/SnoopCompile.jl
@@ -2,6 +2,8 @@ __precompile__()
 
 module SnoopCompile
 
+using Serialization
+
 export
     @snoop,
     @snoop1
@@ -25,27 +27,30 @@ function snoop(filename, commands)
     # addprocs will run the unmodified version of julia, so we
     # launch it as a command.
     code_object = """
-        while !eof(STDIN)
-            eval(Main, deserialize(STDIN))
-        end
-        """
-    in, io = open(`$(Base.julia_cmd()) --eval $code_object`, "w", STDOUT)
-    serialize(in, quote
-        import SnoopCompile
+            using Serialization
+            using Pkg
+            Pkg.activate(abspath("."))
+            while !eof(stdin)
+                Core.eval(Main, deserialize(stdin))
+            end
+            """
+    in_ = open(`$(Base.julia_cmd()) --eval $code_object`, stdout, write=true)
+    serialize(in_, quote
+              import SnoopCompile
     end)
     # Now that the new process knows about SnoopCompile, it can
     # expand the macro in this next expression
-    serialize(in, quote
+    serialize(in_, quote
           SnoopCompile.@snoop1 $filename $commands
     end)
-    close(in)
-    wait(io)
+    # close(in_)
+    # wait(in_)
     println("done.")
     nothing
 end
 
 function split2(str, on)
-    i = search(str, on)
+    i = something(findfirst(isequal(on), str), 0)
     first(i) == 0 && return str, ""
     return (SubString(str, start(str), prevind(str, first(i))),
             SubString(str, nextind(str, last(i))))
@@ -66,11 +71,11 @@ macro snoop1(filename, commands)
     commands = esc(commands)
     return quote
         let io = open($filename, "w")
-            ccall(:jl_dump_compiles, Void, (Ptr{Void},), io.handle)
+            ccall(:jl_dump_compiles, Nothing, (Ptr{Nothing},), io.handle)
             try
                 $commands
             finally
-                ccall(:jl_dump_compiles, Void, (Ptr{Void},), C_NULL)
+                ccall(:jl_dump_compiles, Nothing, (Ptr{Nothing},), C_NULL)
                 close(io)
             end
         end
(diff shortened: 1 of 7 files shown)


timholy:
Thanks! Of the various PRs this looks the most complete.

> `ccall(:jl_generating_output, Cint, ()) == 1 || return nothing`

In older Julia versions, if the package wasn't being pre-compiled this lead to a quick exit. For example if you `include("MyPkg.jl")` rather than `using MyPkg`. This should return 1 only for the process that gets spun up in order to do the precompilation; it should indeed return 0 otherwise. I haven't dug into this myself, are you sure you needed to change this?

src/SnoopCompile.jl
```diff
@@ -25,29 +27,32 @@ function snoop(filename, commands)
...
-        while !eof(STDIN)
-            eval(Main, deserialize(STDIN))
-        end
-        """
-    in, io = open(`$(Base.julia_cmd()) --eval $code_object`, "w", STDOUT)
-    serialize(in, quote
-        import SnoopCompile
+            using Serialization
+            using Pkg
+            Pkg.activate(abspath("."))
```
timholy: If this is necessary can't we just tell people to do this before running `@snoop`?

src/SnoopCompile.jl
```diff
@@ -25,29 +27,32 @@ function snoop(filename, commands)
...
-    wait(io)
+    # close(in_)
+    # wait(in_)
     println("done.")
     nothing
 end
 
 function split2(str, on)
-    i = search(str, on)
+    i = something(findfirst(isequal(on), str), 0)
```
timholy: The more Julian way now would be to just look for `nothing` as the output of `findfirst`. (It's less ambiguous than using a 0 index as "not found" since arrays can now have 0 and negative indices.)

src/SnoopCompile.jl
```diff
@@ -85,6 +90,7 @@ function read(filename)
     data = Vector{String}()
     toplevel = false # workaround for Julia#22538
     for line in eachline(filename)
+        println(line)
```
timholy: leftover?

src/SnoopCompile.jl
```diff
@@ -239,7 +249,7 @@ function write(prefix::AbstractString, pc::Dict)
     for (k, v) in pc
         open(joinpath(prefix, "precompile_$k.jl"), "w") do io
             println(io, "function _precompile_()")
-            println(io, "    ccall(:jl_generating_output, Cint, ()) == 1 || return nothing")
+            println(io, "    ccall(:jl_generating_output, Cint, ()) == 0 || return nothing")
```
timholy: See other comments

src/SnoopCompile.jl
dalum: Oops!

dalum:
Hm.  Maybe the way I'm using snooping is wrong, then.  The specific use I have is for a game/interactive simulation, where I want to compile all the functions, before the user actually interacts.  I thought snoop was pretty handy for this, since it allows me to just run the program, snoop on all the function calls, and then precompile them, rather than depending on some testset, where certain user interactions may be difficult to mimic.  But the `_precompile_()` is not called inside a module that is `used`, but rather run as `julia --project program.jl`.

src/SnoopCompile.jl
dalum: That won't work, since it must be included in the new process that is being spawned.  I am actually a little confused as to the intended difference in purpose between `@snoop` and `@snoop1`?

timholy:
In general the intention was to take the file that it writes as output and `include` it into the project. So the workflow I intended to support was:

- developer writes a package, turns on precompilation (which is now the default)
- developer uses SnoopCompile to analyze what compilation is needed for a particular workflow (e.g., running the tests)
- developer takes the generated `precompile` statements and makes them part of the project, e.g., by `git commit`ing the file and adding `include("precompiles.jl")` as the last line before the module's `end` statement
- the next time someone says `using SomePkg`, precompilation takes longer (not so good) but from that point forward it's faster to use the package (good)

If you just want to force compilation in an interactive setting, you can just "exercise" the required methods. For example if you `include("runtests.jl")` and then do it again, it's usually much faster the second time. (However, not all packages have tests that support re-execution in the same process.)

dalum:
The whole reason that I decided to try snooping was that "exercising" every method being used through testing is too tedious.  But I do think SnoopCompile does exactly the right thing for my use case, except that line, so maybe it could be switched off using some parameter in `write`?

timholy:
Sure, I'm happy to see the scope expand if there's a use case. I see you've handled this very cleaning. I'm playing with your PR a bit now, more in a bit.

timholy:
Note you don't have to address the comments above, since I included them in #19.

This package is hard to review in the abstract, I kind of had to just get my hands on it.

PR #19, opened 2018-08-26
More 0.7/1.0 fixes

timholy:
Builds off #18

### DIFF
--- a/src/SnoopCompile.jl
+++ b/src/SnoopCompile.jl
@@ -2,6 +2,8 @@ __precompile__()
 
 module SnoopCompile
 
+using Serialization
+
 export
     @snoop,
     @snoop1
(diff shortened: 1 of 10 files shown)


src/SnoopCompile.jl
```diff
@@ -25,58 +24,36 @@ function snoop(filename, commands)
...
     # launch it as a command.
     code_object = """
-        while !eof(STDIN)
-            eval(Main, deserialize(STDIN))
+            using Serialization
+            while !eof(stdin)
+                Core.eval(Main, deserialize(stdin))
+            end
+            """
+    process = open(`$(Base.julia_cmd()) --eval $code_object`, stdout, write=true)
```
dalum: Could we maybe allow adding some optional flags like `--project` here?  Maybe just allow passing an iterable of flags to `@snoop`, which will be interpolated into the command.

timholy:
Thanks for adding the patch! I just updated the README and examples (which led to a couple more fixes in the package itself). If you would be willing to include something about the optional flags in the README, I think we can merge and tag this.

timholy:
Bump @dalum. I'm just hoping for 2-3 lines added to the README that show how one might use this. Nothing elaborate, just something that (1) points out this is possible and (2) demonstrates the proper syntax. Then we can merge this and tag it.

dalum:
@timholy Sorry for being late! :frowning: I think it's ready?

As an aside, Simon Danisch pointed out on Slack that the new flag `--trace-compile` basically does the same thing as SnoopCompile (and also that snoop doesn't work on 1.0 ... :wink:)

timholy:
SnoopCompile could basically become the "parcelator" that divides things across packages. Will look into it eventually.

timholy:
Thanks for your help here!

PR #24, opened 2019-03-29
Add the ability to snoop on inference

timholy:
The new feature works on julia v"1.2.0-DEV.564" or higher.


timholy:
Won't merge until at least one Travis worker is new enough to actually test the new functionality.

Issue #29, opened 2019-05-31
Status of SnoopCompile


lkapelevich:
We are trying to understand how to build a user image in Julia and playing with SnoopCompile/PackageCompiler. There are a couple of things that would be nice to clarify. 

1. We get different precompile statements from SnoopCompile/PackageCompiler [example](https://gist.github.com/lkapelevich/e4f539e7da084240a3c49a76554eb0bc) - why is that?
2. Is it feasible to provide a single way to snoop in the future?
3. What is the recommended way to build a user image from SnoopCompile snoop files? Presumably with PackageCompiler, but this is not documented at the moment. (And currently only @SimonDanisch's sd-notomls branch works for unregistered packages, with a bit of manual editing of snoop files)

cc: @SimonDanisch

timholy:
Best guesses:

1. SnoopCompile is careful to launch a new Julia process for the snoop commands. If PackageCompiler isn't, the fact that you already used JLD a bit in the same session means that it will omit the methods that were already compiled. Try again without building your JLD file in the same session.
2. That's more a question for @SimonDanisch since SnoopCompile is the "original."
3. Documented https://github.com/timholy/SnoopCompile.jl#userimgjl

lkapelevich:
Thanks for the responses!

Regarding 3 - Oops I have only tried creating images as in https://docs.julialang.org/en/v1/devdocs/sysimg/index.html and didn't realize we can modify base/uerimg.jl.
I now have a usage question... for the example in the gist, building Julia fails with an `UndefVarError` due to the statement `precompile(Tuple{getfield(JLD, Symbol("##s27#8")), Int, Int, Int, Int, Int})` , is that an issue or am I not including something I should be?

Regarding 1- I didn't mean to build the JLDs in the same session, but I did mess something up and the gist is now updated- I am still curious why the precompile statements are different.

timholy:
> I now have a usage question... for the example in the gist, building Julia fails with an UndefVarError due to the statement precompile(Tuple{getfield(JLD, Symbol("##s27#8")), Int, Int, Int, Int, Int}) , is that an issue or am I not including something I should be?

Are you using the latest SnoopCompile? #27 should have fixed that. If it didn't, check the code behind that change (it was a simple PR), it should give you hints about what else needs to be fixed. (PR appreciated.)

[...]

PR #30, opened 2019-06-06
check that functions in precompile lists are defined

lkapelevich:
Addresses #29 

This works for the [example](https://gist.github.com/lkapelevich/e4f539e7da084240a3c49a76554eb0bc) in the issue, but I'm not sure I have implemented it in the most effective way, or what the best way to test this is. If `func` (https://github.com/timholy/SnoopCompile.jl/blob/master/src/SnoopCompile.jl#L194) looks like 
`:(getfield(mod, Symbol("##...#..")))`, we add a check `isdefined(mod, "##...#..")`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,8 +13,9 @@ julia = "0.7, 1"
 [extras]
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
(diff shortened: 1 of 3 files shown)


timholy:
Sorry I didn't notice this or forgot about it!

I don't see anything objectionable here. Any reason not to merge for now?

lkapelevich:
No problem! I don't have any more changes to make to the code. If you have suggestions for how to test this better, I'm happy to try them.

PR #31, opened 2019-11-19
Tweak `anonrex`

timholy:
Should also get the tests passing on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ os:
 julia:
   - 0.7
   - 1.0
+  - 1.2
   - nightly
 notifications:
   email: false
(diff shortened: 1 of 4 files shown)


yhls:
Did we want to exclude anonymous functions of the form `##something#123`?

timholy:
I'm torn. Generators like `##s27#8` are reproducible until the code changes, maybe it doesn't hurt to keep them?

yhls:
I think they're not reproducible across different versions of Julia, but I guess #30 makes them harmless.

timholy:
There will be a different precompile cache anyway for different versions.

I think I'll go with this and we can consider pruning more methods later.

PR #32, opened 2019-11-20
Get the docstring for snoopi working

timholy:
There's no reason the old one shouldn't have worked, but whatever...


timholy:
https://github.com/JuliaLang/julia/issues/33900

PR #33, opened 2019-11-24
Adding a full real-world example

aminya:
@timholy If you like, you can merge this for a full real-world example for using SnoopCompile

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ _precompile_()
 end # module SomeModule
 ```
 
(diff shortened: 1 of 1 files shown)


timholy:
This is reasonable, but perhaps we should allow multiple packages. (This is not the first real-world user of SnoopCompile.) Can you maybe set it up so that it's a list, with MatLang currently the only entry in the list?

aminya:
@timholy That's a good idea. Done!

timholy:
Thanks!

timholy:
Sorry to have also shaken things up rather a lot since you last looked at this package!

PR #39, opened 2019-12-01
Adding timesum

aminya:
Calculates and prints the total time measured by a snoop macro


src/SnoopCompile.jl
```diff
@@ -71,6 +71,51 @@ if VERSION >= v"1.2.0-DEV.573"
 
 end
 
+"""
+    timesum(snoop)
+
+Calculates and prints the total time measured by a snoop macro
+
+# Examples
+To benchmark the loadTime and RunTime of a package:
```
timholy: I worry that this is misleading. Much of the load time is spent inserting methods into method tables (https://discourse.julialang.org/t/how-to-debug-compilation-time-and-loading-time-of-packages/31583/4?u=tim.holy); for run time, keep in mind that `@snoopi` only measures inference time, and the run time of the tests might depend on computation time as well as inference time.

It would be accurate to say that these measure the *inference time* from loading the package and running the tests.

src/SnoopCompile.jl
```diff
@@ -71,6 +71,51 @@ if VERSION >= v"1.2.0-DEV.573"
...
+    using ColorTypes
+    include(joinpath(dirname(dirname(pathof(ColorTypes))),"test","runtests.jl"))
+end
+
+timesum(runSnoop)
+```
+"""
+export timesum
+function timesum(snoop::Vector{Tuple{Float64, Core.MethodInstance}})
+    timeSum = 0
```
timholy:
```suggestion
    timeSum = 0.0
```

src/SnoopCompile.jl
```diff
@@ -71,6 +71,51 @@ if VERSION >= v"1.2.0-DEV.573"
...
+    for x in snoop
+        timeSum+=x[1]
+    end
+    println(timeSum)
```
timholy:
```suggestion
```
If you return it, I don't think you have to print it. (The user can choose that for him/herself.)

src/SnoopCompile.jl
aminya: Here I meant inferring, but I need to do clarification. 

So for general compiling time benchmarking, do you recommend `snoopi` or `snoopc`?

timholy: They measure very different things. The reason `@snoopi` is more aligned with the purpose of this package is that currently Julia can only cache inference results; hence inference time, rather than compile time, is the benchmark that you can expect to see reduced by adding precompile statements.

aminya:
I cannot design a test that checks the value returned by timesum. I think the function is very straightforward and easy. It doesn't need an exhaustive test. I just added a syntax and type checking test.

timholy:
Need to fix all the travis failures before this can be merged. You're using `@snoopi` on all Julia versions, but it's supported only on Julia 1.2 and higher. Also, `ColorTypes` includes a method-ambiguity check, and by running its tests after loading JLD in the same process, you trigger a failure of the ambiguity check. Perhaps pick a different package, or write a mini-package here (in the `test/` directory) specifically for testing.

aminya:
I don't know what else to do. I might close the PR. It doesn't worth it.

timholy:
You're probably not "protecting" the `@snoopi` macro, e.g., by putting it in a string. (Macro expansion is performed at lowering time, so having it in an `if` block doesn't help.) But if you prefer to close this that's also fine; `sum(first, tinf)` should work, no?

aminya:
I found `@static` macro that allows writing tests for different Julia versions.

timholy:
The Windows test failures will be fixed by #48. The harder question is whether to drop 0.7. I suppose we can do that. But I do want to make sure that we support many Julia versions given concerns like #45; if anything, the testing across versions needs to become more stringent, not less.

aminya:
I think by using `@static` we can write tests easily for different versions. So using that dropping 0.7 doesn't make developing easier any more.
If you think a lot things are going to change then we can drop 0.7.

Many packages don't support 0.7, and that is totally fair.

timholy:
I've fixed the merge conflicts, squashed, and submitted this as #51

PR #43, opened 2019-12-03
Encapsulate ccalls to enhance interpreter usage

timholy:
If Julia compiles the returned expression, we will "miss" the inference time because the compilation happens before we turn snooping on.

The logic [here](https://github.com/JuliaLang/julia/blob/84e839227f7762b6dfa0821247cd132cf15e3581/src/toplevel.c#L793-L815) indicates that the `ccall`s could force Julia to use the compiler, so move these inside a function.

### DIFF
--- a/src/SnoopCompile.jl
+++ b/src/SnoopCompile.jl
@@ -23,6 +23,9 @@ if VERSION >= v"1.2.0-DEV.573"
         return ret
     end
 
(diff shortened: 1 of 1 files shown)


timholy:
Oddly, the two builds used different versions of nightly. Passes on the more recent, so this is good to merge.

PR #48, opened 2019-12-05
Invariant naming of generators

timholy:
@chriselrod, can you test to see if this fixed your test cases?


chriselrod:
I can confirm that my testcase now works.
Sorry that the code I provided didn't actually work, but thank you for fixing this!

Issue #37, opened 2019-12-01
Github action to automatically update precompile files


aminya:
Automatic repository updating using Github actions for uploading the files generated by snoopCompile.jl script

Similar to https://github.com/julia-actions/julia-uploadcodecov or https://github.com/julia-actions/julia-docdeploy

timholy:
Cool idea. These seem to require JavaScript, a language I don't speak, so I'd rely on external contributors for such bots.

aminya:
Also, I found this push bot: https://github.com/ad-m/github-push-action which does exactly what we want. We just need to run the Julia file before this.

aminya:
Probably @davidanthoff can help us with this

aminya:
I made a workflow file that does what we want! 😃 
https://github.com/juliamatlab/MatLang/blob/master/.github/workflows/SnoopCompile.yml

davidanthoff:
Cool! Once you guys have tried it out for a while I'd be happy to create a proper github action and integrate it with https://github.com/davidanthoff/PkgButlerEngine.jl.

aminya:
I get this error: https://github.com/aminya/AcuteML.jl/commit/141099db133d89c147f94779d06a20cdbed6be65/checks?check_suite_id=337167681#step:7:5
Do you know what the reason is?
```julia
Launching new julia process to run commands...
ERROR: LoadError: could not open file /home/runner/work/AcuteML.jl/AcuteML.jl/deps/SnoopCompile/precompile/precompile_AcuteML.jl
Stacktrace:
 [1] include at ./boot.jl:328 [inlined]
 [2] include_relative(::Module, ::String) at ./loading.jl:1105
 [3] include at ./Base.jl:31 [inlined]
 [4] include(::String) at /home/runner/work/AcuteML.jl/AcuteML.jl/src/AcuteML.jl:1
 [5] top-level scope at /home/runner/work/AcuteML.jl/AcuteML.jl/src/AcuteML.jl:1055
 [6] include at ./boot.jl:328 [inlined]
 [7] include_relative(::Module, ::String) at ./loading.jl:1105
 [8] include(::Module, ::String) at ./Base.jl:31
 [9] top-level scope at none:2
 [10] eval at ./boot.jl:330 [inlined]
 [11] eval(::Expr) at ./client.jl:425
 [12] top-level scope at ./none:3
in expression starting at /home/runner/work/AcuteML.jl/AcuteML.jl/src/AcuteML.jl:1055
[...]

PR #49, opened 2019-12-06
Utilities for SnoopCompile Github actions

aminya:
Adds support for https://github.com/timholy/SnoopCompile.jl/issues/37

These are some utilities that I will be using for the SnoopCompile bot. I think the best place for these functions is the SnoopCompile itself.


aminya:
I cannot get the `snoopiBot` macro working. It works with directly running the code.

Here is the error I get:
https://github.com/juliamatlab/MatLang/runs/345441347#step:7:4
It says key `:MatLang` is not defined. I am not sure why the key isn't interpolated correctly in the macro.

To test the macro one can run this:
https://github.com/juliamatlab/MatLang/blob/master/deps/SnoopCompile/snoopCompile.jl

aminya:
I think I found the issue. The problem seems to be my macro doesn't evaluate another inner macro.

Do I need to eval it manually?
```
using Test
macro foo(in::String)
      quote
          b = @test 1 == 1
          symb = Symbol($in)
          d = Dict(symb => b)
          dsymb = d[symb]
      end
end
@foo "bar"
```

aminya:
The code is tested and it works! 
https://github.com/juliamatlab/MatLang/commit/81ff83a773c1d3c3ee71711da8e7b82d50d85cdb/checks?check_suite_id=354813377

KristofferC:
Not directly related to this PR but https://docs.julialang.org/en/v1/manual/style-guide/index.html#Use-naming-conventions-consistent-with-Julia-base/-1 is good reading when contributing to packages, as well as looking around a bit in the code base of the package itself to see how things are done.

aminya:
Fixed a bug that could happen in some cases.

Now the run works without any issues:
https://github.com/aminya/AcuteML.jl/commit/f259cd00ee01e124a8c34de012702b16e89616de/checks?check_suite_id=376910459

The pull request is created successfully:
https://github.com/aminya/AcuteML.jl/pull/42

aminya:
Created a workflow for adding test dependencies.

Now the run works without any issues for [Juno.jl](https://github.com/JunoLab/Juno.jl/pull/437):
https://github.com/aminya/Juno.jl/commit/20368ee2b7624425e65c0a3940bdf1d5eedac531/checks?check_suite_id=377037229

The pull request is created successfully:
https://github.com/aminya/Juno.jl/pull/3

----------------------------------------------------------

However, for [Atom](https://github.com/JunoLab/Atom.jl/pull/225), runtests.jl uses `@__DIR__` internally for path completion test which makes the test fail. Because the test runs from our macro,  `@__DIR__`  gives the wrong answer.
https://github.com/aminya/Atom.jl/runs/366285209

aminya:
Here are the PRs after using Julia nightly:
MatLang:
https://github.com/juliamatlab/MatLang/pull/27/files
Atom:
https://github.com/aminya/Atom.jl/pull/5
Juno:
https://github.com/aminya/Juno.jl/pull/4

SnoopCompile automatically handles different version differences.

aminya:
I added a benchmark bot too.
 
Running an online benchmark enhances:

- MatLang examples infer time by 35% 

https://github.com/juliamatlab/MatLang/commit/5d2a9b2b53dee15db6f208b0dd112043a012eb6b/checks?check_suite_id=378843420

- Juno tests infer time by 25%

https://github.com/aminya/Juno.jl/commit/416911b2befa692215893bc74fa2ea3fd0383b5e/checks?check_suite_id=378856765

timholy:
Let me know when you think this is ready and I'll give it a review.

aminya:
@timholy The PR is ready. I would like to have some feature in the parcel that I have created an issue for it: https://github.com/timholy/SnoopCompile.jl/issues/56

When it is added, I would add it to BotConfig too.

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+        os: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.julia-version }}
+      - name: Install dependencies
+        run: julia --project=@. -e 'using Pkg; Pkg.instantiate();'
+      - name : Add SnoopCompile and current package
+        run: julia -e 'using Pkg; Pkg.add(PackageSpec(url = "https://github.com/aminya/SnoopCompile.jl", rev ="packageSnooper")); Pkg.develop(PackageSpec(; path=pwd()));'
```
timholy: This link will have to change

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+  using MatLang
+  examplePath = joinpath(dirname(dirname(pathof(MatLang))), "examples")
+  include(joinpath(examplePath,"Language_Fundamentals", "usage_Entering_Commands.jl"))
+  include(joinpath(examplePath,"Language_Fundamentals", "usage_Matrices_and_Arrays.jl"))
+  include(joinpath(examplePath,"Language_Fundamentals", "Data_Types", "usage_Numeric_Types.jl"))
+end
+```
+[Ref]( https://github.com/juliamatlab/MatLang/blob/master/deps/SnoopCompile/snoopCompile.jl)
+
+or If you do not have additional examples, you can use your runtests.jl file:
```
timholy:
```suggestion
or if you do not have additional examples, you can use your runtests.jl file:
```

src/bot.jl
```diff
@@ -0,0 +1,30 @@
+export precompileActivator, precompileDeactivator, precompilePather, @snoopiBot, @snoopiBenchBot, BotConfig
```
timholy: Please respect the standard Julia naming conventions. See https://github.com/timholy/SnoopCompile.jl/pull/49#issuecomment-565013426.

src/bot.jl
```diff
@@ -0,0 +1,30 @@
...
+UStrings = Union{AbstractString,Regex,AbstractChar}
+################################################################
+"""
+    BotConfig
+
+Config object that holds the options and configuration for the snoopCompile bot. This object is fed to the `@snoopiBot`.
+
+# Arguments:
+- packageName::String
+- subst::Vector{Pair{UStrings, UStrings}} : to replace a packages precompile setences with another's package like ["ImageTest" => "Images"]
```
timholy: `UStrings` isn't documented

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+To get the path of precompile_packageName.jl file
+
+Written exclusively for SnoopCompile Github actions.
+# Examples
+```julia
+precompilePath, precompileFolder = precompilePather("MatLang")
+```
+"""
+function precompilePather(packageName::String)
+    return "\"../deps/SnoopCompile/precompile/precompile_$packageName.jl\"",
```
timholy: `parcel` includes the package name as part of the file name, but when I copy the files to a project's `src/` directory I often strip the package name off and just call it `precompile.jl`. Is it possible to detect this case too?

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+    c1 = Regex("#[^\\S\\r\\n]{0,}include\\($(precompilePath)\\)")
+    c2 = r"#\s{0,}_precompile_\(\)"
+    a1 = "include($precompilePath)"
+    a2 = "_precompile_()"
+    return c1, c2, a1, a2
+end
+################################################################
+
+"""
+    precompileActivator(packagePath, precompilePath)
```
timholy: What, exactly, is `packagePath` supposed to be? The name of a file or a directory? Is it always the top-level file? What if the `_precompile_()` isn't in that file but some other file?

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+
+Activates precompile of a package by adding or uncommenting include() of *.jl file generated by SnoopCompile and _precompile_().
+
+Written exclusively for SnoopCompile Github actions.
+"""
+function precompileActivator(packagePath::String, precompilePath::String)
+
+    file = open(packagePath,"r")
+    packageText = Base.read(file, String)
+    close(file)
```
timholy: You can replace this whole block with `package_text = read(filename, String)`. This applies to other places too, but I won't comment there.

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+        packageEdited = foldl(replace,
+                     (
+                      c1 => a1,
+                      c2 => a2,
+                     ),
+                     init = packageText)
+
+                     file = open(packagePath,"w")
+                     Base.write(file, packageEdited)
+                     close(file)
```
timholy: This can also be `write(filename, package_edited)` (also applies elsewhere)

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+    precompileActivator(packagePath, precompilePath)
+
+Activates precompile of a package by adding or uncommenting include() of *.jl file generated by SnoopCompile and _precompile_().
+
+Written exclusively for SnoopCompile Github actions.
+"""
+function precompileActivator(packagePath::String, precompilePath::String)
+
+    file = open(packagePath,"r")
+    packageText = Base.read(file, String)
```
timholy: This variable name kind of implies this is the entire package definition, but many packages have many source files.

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+                     file = open(packagePath,"w")
+                     Base.write(file, packageEdited)
+                     close(file)
+        println("precompile is activated")
+    elseif available
+        # do nothing
+        println("precompile is already activated")
+    else
+        # TODO: add code automatiaclly
+        error(""" add the following codes into your package:
```
timholy: Where? What if the package has many files in it? Can it occur before any of the code definitions? (Answer: no.) Perhaps link to this section of the docs: https://timholy.github.io/SnoopCompile.jl/stable/snoopi/#auto-1, which spells out in greater detail exactly where to put this.

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,120 @@
...
+         #include($precompilePath)
+         #_precompile_()
+         """)
+    end
```
timholy: I worry that routine editing of a package might result in a file with the following content:
```julia
# include("precompile.jl")
# _precompile_()

# some other stuff

include("precompile.jl")
_precompile_()
```
Obviously this is a user error, but what happens in this case?

src/bot/snoopiBenchBot.jl
```diff
@@ -0,0 +1,132 @@
...
+```
+
+Benchmarking the example infer time
+```julia
+println("examples infer benchmark")
+
+@snoopiBenchBot "MatLang" begin
+    using MatLang
+    examplePath = joinpath(dirname(dirname(pathof(MatLang))), "examples")
+    # include(joinpath(examplePath,"Language_Fundamentals", "usage_Entering_Commands.jl"))
```
timholy: What's the purpose of having this commented line?

src/bot/snoopiBenchBot.jl
```diff
@@ -0,0 +1,132 @@
...
+        return sum(first, snoop)
+    end
+end
+
+################################################################
+"""
+    @snoopiBenchBot(packageName::String, snoopScript::Expr)
+    @snoopiBenchBot(packageName::String)
+
+Performs an infertime benchmark by activating and deactivating the __precompile__()
```
timholy: 2-underscores is very different from 1:
```julia
help?> __precompile__
search: __precompile__

  __precompile__(isprecompilable::Bool)

  Specify whether the file calling this function is precompilable, defaulting
  to true. If a module or file is not safely precompilable, it should call
  __precompile__(false) in order to throw an error if Julia attempts to
  precompile it.
```
Maybe I should have not chosen something so easily confused :disappointed:

src/bot/snoopiBot.jl
```diff
@@ -0,0 +1,109 @@
...
+
+        ### Log the compiles
+        data = @snoopi begin
+            $(esc(snoopScript))
+        end
+
+        ################################################################
+        ### Parse the compiles and generate precompilation scripts
+        pc = SnoopCompile.parcel(data, subst = $subst, blacklist = $blacklist)
+        onlypackage = Dict( packageSym => sort(pc[packageSym]) )
```
timholy: This `sort` isn't necessary anymore I think

src/bot/snoopiBot.jl
```diff
@@ -0,0 +1,109 @@
...
+end
+
+macro snoopiBot(packageName::String, snoopScript)
+    config = BotConfig(packageName)
+    return quote
+        @snoopiBot $config $(esc(snoopScript))
+    end
+end
+macro snoopiBot(configExpr, snoopScript)
+    config = eval(configExpr)
```
timholy: What's this about? Can't we just pass back the expression and have it eval normally?

test/bot/deactivated.jl
```diff
@@ -0,0 +1,2 @@
+# include("../deps/SnoopCompile/precompile/precompile_TestPackage.jl")
```
timholy: Rather than committing this to the project, it might be best to have the `"precompileActivator"` testset write it, most likely to some tempfile. That way, if the tests get interrupted between "activator" and "deactivator", the directories won't be left in an inconsistent state.

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
+# SnoopCompile Bot
```
timholy:
```suggestion
# SnoopCompile Bot (EXPERIMENTAL)
```

test/bot/bot.jl
```diff
@@ -0,0 +1,59 @@
...
+
+        @snoopiBenchBot "MatLang" begin
+            using MatLang
+            examplePath = joinpath(dirname(dirname(pathof(MatLang))), "examples")
+            # include(joinpath(examplePath,"Language_Fundamentals", "usage_Entering_Commands.jl"))
+            include(joinpath(examplePath,"Language_Fundamentals", "usage_Matrices_and_Arrays.jl"))
+            include(joinpath(examplePath,"Language_Fundamentals", "Data_Types", "usage_Numeric_Types.jl"))
+        end
+    end
+    =#
```
timholy: If this path issue is fixable it would be reassuring to have this functionality tested.

docs/src/bot.md
KristofferC: Perhaps use `julia --project` and remove the `Pkg.develop` call.

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+        julia-version: ['nightly']
+        julia-arch: [x64]
+        os: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.julia-version }}
+      - name: Install dependencies
+        run: julia --project=@. -e 'using Pkg; Pkg.instantiate();'
```
KristofferC: I think this can just be 

```suggestion
        run: julia --project -e 'using Pkg; Pkg.instantiate();'
```

src/bot/botutils.jl
```diff
@@ -0,0 +1,19 @@
...
+Should be removed once Pkg allows adding test dependencies to the current environment
+
+Used in Github Action workflow yaml file
+"""
+function addtestdep()
+    local testToml
+    try
+        testToml = Pkg.Types.parse_toml("test/Test.toml")
+    catch
+        error("please add a Test.toml to the /test directory for test dependencies")
```
KristofferC: This seems like it is introducing a hole new way to give test dependencies. What is the use case?

Also, instead of a try catch, it seems that a `isfile` would work just as well considering the error message.

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+      - name: Running Benchmark
+        run: julia --project=@. -e 'include("deps/SnoopCompile/snoopBenchmark.jl")'
+
+      # https://github.com/marketplace/actions/create-pull-request
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v2-beta
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update precompile_*.jl file
+          committer: Amin Yahyaabadi <aminyahyaabadi74@gmail.com>
```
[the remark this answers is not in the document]
aminya: I had a comment for it below the code.
https://github.com/aminya/SnoopCompile.jl/blame/packageSnooper/docs/src/bot.md#L71

But you are right it makes sense to move the comment inside the code itself.

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+on:
+  - push
+
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: ['nightly']
```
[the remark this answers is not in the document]
aminya: To use the latest SnoopCompile feature that handles different Julia versions, and also removes keyword arguments. If we can provide this feature in a released Julia, we can change this.

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+```yaml
+# - name: Install Test dependencies
+#  run: julia -e 'using SnoopCompile; SnoopCompile.addtestdep()'
+```
+
+For example for MatLang package:
+
+[Link](https://github.com/juliamatlab/MatLang/blob/master/.github/workflows/SnoopCompile.yml)
+
+Change `committer` to your name and your email.
```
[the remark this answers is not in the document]
aminya: Fixed in https://github.com/timholy/SnoopCompile.jl/pull/49/commits/582d5740506f03e08a65f2a3b69e5250cd47b931

src/bot/snoopiBot.jl
aminya: I use this line to get `BotConfig` type and dispatch on it for the macro. Because macros can't detect types, I have to eval it.

src/bot/snoopiBot.jl
aminya: So does parcel sort the signatures already? 

The reason I have this is to avoid creating pull requests that just change the order of lines.

src/bot/snoopiBenchBot.jl
aminya: I needed the path of the test package so I can cd to it, but without loading the package. So I decided to have an external Julia process.

src/bot.jl
aminya: I don't know what I should do regarding this. What are your suggested names?

src/bot.jl
aminya: It is documented below this section

src/bot/precomileInclude.jl
aminya: No. Now the precompile signatures should be kept in this particular path, otherwise the bot will not work.

src/bot/snoopiBot.jl
timholy: Actually, only `write` does this: https://github.com/timholy/SnoopCompile.jl/blob/95103414dc870dfab0d46118dcb9c050b038c5fe/src/write.jl#L38

So keep it if you need it.

src/bot/snoopiBenchBot.jl
aminya: Ahh. My bad. I didn't pay enough attention. 😄

src/bot/precomileInclude.jl
aminya: It is the main package file that has the `_precompile_`

src/bot/precomileInclude.jl
aminya: https://github.com/timholy/SnoopCompile.jl/pull/49#discussion_r362505975 
Should I change the variable name? 

It is the same as `pathof`

aminya:
@KristofferC @timholy regarding `addtestdep` there is no way to add the tests deps to the current environment. So I wrote my own way. If there is some way in Pkg itself, I would be happy to remove it.

I asked this a couple of times in the Slack group. https://julialang.slack.com/archives/C67EFTEF3/p1577556453102300

There is an issue for it without any solution
https://github.com/JuliaLang/Pkg.jl/issues/1008

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Update precompile_*.jl file
+          committer: Amin Yahyaabadi <aminyahyaabadi74@gmail.com>
+          title: '[AUTO] Update precompile_*.jl file'
+          labels: SnoopCompile
+          branch: create-pull-request/SnoopCompile
+      - name: Check output environment variable
+        run: echo "Pull Request Number - ${{ env.PULL_REQUEST_NUMBER }}"
+```
+If your examples or tests have dependencies, you should add a `Test.toml` to your test folder.
```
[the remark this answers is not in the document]
aminya: For Julia 1.2 and more people should add a Project.toml to their test folder. So I saw this somehow consistent with that decision. Related: https://github.com/timholy/SnoopCompile.jl/pull/49#issuecomment-570237787

docs/src/bot.md
```diff
@@ -0,0 +1,148 @@
...
+println("loading infer benchmark")
+
+@snoopiBenchBot "MatLang" using MatLang
+```
+
+Benchmarking the example infer time
+```julia
+println("examples infer benchmark")
+
+@snoopiBenchBot "MatLang" begin
```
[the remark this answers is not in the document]
aminya: Yes definitely. Any of these codes can be used offline too. The only thing that enables doing this online is the yaml file. 

So do you think I shuold add some text to doc to clarify that these macros can be used offline too?

docs/src/bot.md
aminya: I want the package to be available globally. Because the `snoopiBot` runs another Julia process.

docs/src/bot.md
aminya: are these the same? 🤔

KristofferC: No argument to `--project?  should use the project file in the current (or parent of current) working directory.

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,114 @@
+
```
KristofferC: This filename has a typo.

src/bot/precomileInclude.jl
```diff
@@ -0,0 +1,114 @@
...
+To get the path of precompile_packageName.jl file
+
+Written exclusively for SnoopCompile Github actions.
+# Examples
+```julia
+precompilePath, precompileFolder = precompilePather("MatLang")
+```
+"""
+function precompilePather(packageName::String)
+    return "\"../deps/SnoopCompile/precompile/precompile_$packageName.jl\"",
```
KristofferC: Should this be something like `joinpath(@__DIR__, "..", "deps", ...)` to be independent of the current working directory?

src/bot.jl
KristofferC: Just some notes,

- variables should be like `precompile_activator` (same with file names)
- `fooBar` type of capitalization is never used
- `FooBar` should be used for structs (this seems to already be the case)

src/bot.jl
```diff
@@ -0,0 +1,30 @@
...
+
+UStrings = Union{AbstractString,Regex,AbstractChar}
+################################################################
+"""
+    BotConfig
+
+Config object that holds the options and configuration for the snoopCompile bot. This object is fed to the `@snoopiBot`.
+
+# Arguments:
+- packageName::String
```
KristofferC: Needs code quoting in a bunch of places.

docs/src/bot.md
```diff
@@ -0,0 +1,146 @@
...
+    examplePath = joinpath(dirname(dirname(pathof(MatLang))), "examples")
+    # include(joinpath(examplePath,"Language_Fundamentals", "usage_Entering_Commands.jl"))
+    include(joinpath(examplePath,"Language_Fundamentals", "usage_Matrices_and_Arrays.jl"))
+    include(joinpath(examplePath,"Language_Fundamentals", "Data_Types", "usage_Numeric_Types.jl"))
+end
+```
+
+Benchmarking the tests:
+```julia
+@snoopiBenchBot "MatLang"
```
KristofferC: What does "benchmarking the tests" mean and how is it related to collecting precompile files?

src/bot.jl
```diff
@@ -0,0 +1,30 @@
+export precompileActivator, precompileDeactivator, precompilePather, @snoopiBot, @snoopiBenchBot, BotConfig
+
+UStrings = Union{AbstractString,Regex,AbstractChar}
```
KristofferC: should be `const UStrings = ...`

src/bot/precomileInclude.jl
timholy: That's what I expect too, but it might be necessary to state this in the docs as a requirement.

docs/src/bot.md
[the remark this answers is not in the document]
timholy: If they can be used outside of the bot, it's confusing to have `Bot` in the name.

docs/src/bot.md
aminya: It is benchmarking the infer time of the tests

src/bot/precomileInclude.jl
timholy: In this case you can just document it. If it's always `pathof(MyPkg)` then is it feasible to pass `MyPkg` as an argument and call `pathof` here? That would ensure you always get the right answer.

aminya:
I will address the rest later. After that I will squash these edits to make the history clean again.

src/bot/precomileInclude.jl
aminya: We should use something like
```julia
packagePath = Base.read(`cmd /c julia -e 'import MatLang; print(pathof(MatLang))'`, String)
```
if we want to calculate pathof automatically.

Anyways these functions are internal to the bot and are not supposed to be used directly.

aminya:
@KristofferC @timholy response to a couple of comments:
https://github.com/timholy/SnoopCompile.jl/pull/49#discussion_r362444307
https://github.com/timholy/SnoopCompile.jl/pull/49#discussion_r362514505
https://github.com/timholy/SnoopCompile.jl/pull/49#discussion_r362442292

I am assuming that the users of this package are following the rules I set and specified in [the documentation](https://github.com/aminya/SnoopCompile.jl/blob/packageSnooper/docs/src/bot.md), and so the code works for these situations. 

Handling all the other possibilities does not have much value.

---------------------------------------------------
For example:

To make the whole code independent of the current working directory and where packages files are placed a couple of things should be done.
For example to get the path of the package we should do:

```julia
packagePath = Base.read(`cmd /c julia -e 'import MatLang; print(pathof(MatLang))'`, String)
```
instead of https://github.com/aminya/SnoopCompile.jl/blob/32d36d1e9a2bc3d22fa90afe3652e2f404689e5d/src/bot/snoopiBot.jl#L30
```julia
packagePath = joinpath(pwd(),"src","$packageName.jl")
```
---------------------------------------------------

or for `precompile_$packageName.jl`, some script can search in the package folder. 

-------------------------------------
or if the user decides to have multiple precompile inclusion by mistake, such as:

```julia
# include("precompile.jl")
# _precompile_()

# some other stuff

include("precompile.jl")
_precompile_()
```
The code will not work.

aminya:
I squashed commits that addressed review comments. If you approve, this can be merged.

aminya:
> Are you planning on maintaining this code long-term? I have made a suggestion to mark it as "EXPERIMENTAL" because I already maintain too much code and do not want to be on the hook for this.

I can maintain the code. Should I remove "EXPERIMENTAL"?

aminya:
@timholy This is ready. Should I do anything else?

timholy:
Let's do it! Thanks for a huge contribution!

timholy:
Version 1.2 is out: https://github.com/JuliaRegistries/General/pull/7894

Care to do the honors of writing a discourse post?

aminya:
> 
> 
> Version 1.2 is out: [JuliaRegistries/General#7894](https://github.com/JuliaRegistries/General/pull/7894)
> 
> Care to do the honors of writing a discourse post?

Awesome! I would be happy to write a post 🚀

src/bot/snoopiBot.jl
```diff
@@ -0,0 +1,109 @@
...
+@snoopiBot "MatLang" begin
+  using MatLang
+  examplePath = joinpath(dirname(dirname(pathof(MatLang))), "examples")
+  include(joinpath(examplePath,"Language_Fundamentals", "usage_Entering_Commands.jl"))
+  include(joinpath(examplePath,"Language_Fundamentals", "usage_Matrices_and_Arrays.jl"))
+  include(joinpath(examplePath,"Language_Fundamentals", "Data_Types", "usage_Numeric_Types.jl"))
+end
+```
+"""
+macro snoopiBot(config::BotConfig, snoopScript)
```
KristofferC: Macro names like `@snoopiBot`, `@snoopiBench` should ideally also respect the style guide line. For someone coming from Julia, the capitalization will look quite weird (almost like it is constructing a type, but not really). Lower Camel Case is never used in Julia

timholy: Yeah, I probably should have insisted on changing these, I just felt guilty about letting this sit for so long. @aminya, this has come up so often that I think you really just need to sit down and carefully read the entire style guide and start embracing it in your code.

aminya: Sorry about that, I have started to use `_` instead of capitalization for functions.

As a solution, we can create aliases if you want to but keep the old function names just for compatibility.

```
export snoopi_bot, snoopi_bench, snoopiBot, snoopiBench 
const snoopiBot = snoopi_bot
const snoopiBench = snoopi_bench
```

timholy: You can probably use `Base.@deprecate_binding` to issue a warning about the old names.

aminya: A patch was created 
https://github.com/timholy/SnoopCompile.jl/pull/66

PR #51, opened 2019-12-07
Add timesum

timholy:
This is #39 with the merge conflicts fixed. @aminya, please see if this seems OK.


timholy:
BTW, I had to drop 0.7 because the error occurs at `Pkg` installation, not when the tests are run. (MatLang is not installable on 0.7.)

aminya:
It looks good to me. Thanks!

Issue #63, opened 2020-01-14
blacklist not working for parcel_snoopi


aminya:
https://github.com/timholy/SnoopCompile.jl/blob/6ad1d8c7c4f7508b55bfe2345ebd58abfa0b60fb/src/parcel_snoopi.jl#L197

PR #64, opened 2020-01-14
blacklist remover

aminya:
Fixes https://github.com/timholy/SnoopCompile.jl/issues/63

### DIFF
--- a/src/parcel_snoopi.jl
+++ b/src/parcel_snoopi.jl
@@ -193,7 +193,6 @@ function handle_kwbody(topmod::Module, m::Method, paramrepr, tt, fstr="fbody")
     return nothing
 end
 
-
 function parcel(tinf::AbstractVector{Tuple{Float64,Core.MethodInstance}}; subst=Vector{Pair{String, String}}(), blacklist=String[])
     pc = Dict{Symbol, Vector{String}}()      # output
     modgens = Dict{Module, Vector{Method}}() # methods with generators in a module
@@ -229,11 +228,7 @@ function parcel(tinf::AbstractVector{Tuple{Float64,Core.MethodInstance}}; subst=
         paramrepr = map(T->reprcontext(topmod, T), Iterators.drop(tt.parameters, 1))  # all the rest of the args
 
         # blacklist remover
-        for (iLine, line) in enumerate(pc[topmodname])
-          if any(occursin.(blacklist, line))
-            deleteat!(pc[topmodname], iLine)
-          end
-        end
+        pc[topmodname] = blacklist_remover(blacklist, pc[topmodname])
 
         if any(str->occursin('#', str), paramrepr)
             @debug "Skipping $tt due to argument types having anonymous bindings"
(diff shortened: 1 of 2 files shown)


src/parcel_snoopi.jl
```diff
@@ -227,6 +227,14 @@ function parcel(tinf::AbstractVector{Tuple{Float64,Core.MethodInstance}}; subst=
         # Use special care with keyword functions, anonymous functions
         p = tt.parameters[1]   # the portion of the signature related to the function itself
         paramrepr = map(T->reprcontext(topmod, T), Iterators.drop(tt.parameters, 1))  # all the rest of the args
+
+        # blacklist remover
+        for (iLine, line) in enumerate(pc[topmodname])
+          if any(occursin.(blacklist, line))
+            deleteat!(pc[topmodname], iLine)
```
timholy: This will affect later iterations in the enumeration. Demo:
```julia
julia> list = [1, 2, 3, 3, 4, 5]
6-element Array{Int64,1}:
 1
 2
 3
 3
 4
 5

julia> function remove3!(list)
           for (i, v) in enumerate(list)
               v == 3 && deleteat!(list, i)
           end
           return list
       end
remove3! (generic function with 1 method)

julia> remove3!(list)
5-element Array{Int64,1}:
 1
 2
 3
 4
 5
```
Only the first 3 is removed because the index advanced past the one that followed it.

Instead, accumulate all the indexes you want to delete to temporary storage and delete them at the conclusion of the loop.

aminya: Ahhh. That's right. :smile: I didn't notice that. I will fix it

src/parcel_snoopi.jl
```diff
@@ -297,3 +300,25 @@ function parcel(tinf::AbstractVector{Tuple{Float64,Core.MethodInstance}}; subst=
...
+
+# Examples
+```julia
+blacklist = ["hi","bye"]
+pcI = ["good","bad","hi","bye","no"]
+
+blacklist_remover(blacklist, pcI)
+```
+"""
+function blacklist_remover(blacklist, pcI)
```
timholy: Please swap the order of arguments and call this `blacklist_remover!`.

https://docs.julialang.org/en/v1.3/manual/style-guide/ (Seem familiar? :stuck_out_tongue_winking_eye: )

src/parcel_snoopi.jl
```diff
@@ -297,3 +300,25 @@ function parcel(tinf::AbstractVector{Tuple{Float64,Core.MethodInstance}}; subst=
...
+# Examples
+```julia
+blacklist = ["hi","bye"]
+pcI = ["good","bad","hi","bye","no"]
+
+SnoopCompile.blacklist_remover(blacklist, pcI)
+```
+"""
+function blacklist_remover(blacklist, pcI)
+    idx = Vector{Int64}(undef, 0)
```
timholy: Should be `Int`, not `Int64`. And shorter: `idx = Int[]`.

aminya:
Ready to get merged

PR #82, opened 2020-05-24
Fix macro name

PallHaraldsson:
[ci skip]

### DIFF
--- a/docs/src/snoopc.md
+++ b/docs/src/snoopc.md
@@ -45,7 +45,7 @@ There are more complete example illustrating potential options in the `examples/
 
 When calling the `@snoopc` macro, a new julia process is spawned using the function `Base.julia_cmd()`.
 Advanced users may want to tweak the flags passed to this process to suit specific needs.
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
I think that was a typo, couldn't locate `@snoop` (search isn't too helpful for "`@snoop(`", at least when searching for non-existent), and I think and hope I got the right replacement.

I trying to understand this package... and maybe docs could be clearer, and I could make a PR.

PallHaraldsson:
I got:

```
julia> inf_timing = @snoopi sum(a)
Tuple{Float64,Core.MethodInstance}[]
```

and then realized nothing wrong with docs, rather I was running like:
```
$ ~/julia-1.6.0-DEV-8f512f3f6d/bin/julia -O0 --compile=min --startup-file=no
```

and I at least learned something, `--compile=min` disables inference. I'm just not sure we should complicate the docs explaining undocumened feature. Should it be undocumened? I made a PR at Julia do change that.

timholy:
Thanks! I'd be delighted to have more doc improvements.

PR #88, opened 2020-06-15
GoodPath on pathof

aminya:


aminya:
@timholy This should fix the path issue.

PR #92, opened 2020-06-15
Invalidation show

aminya:
- [x] Use `ing` for reason
- [x] Use `printstyled` to print the code for methods with color

### DIFF
--- a/examples/invalidations_blog.jl
+++ b/examples/invalidations_blog.jl
@@ -1,5 +1,5 @@
 using SnoopCompile
-using SnoopCompile: countchildren
+import SnoopCompile: countchildren
 
 function hastv(typ)
     isa(typ, UnionAll) && return true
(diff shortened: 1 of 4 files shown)


examples/invalidations_blog.jl
```diff
@@ -1,5 +1,5 @@
 using SnoopCompile
-using SnoopCompile: countchildren
+import SnoopCompile: countchildren
```
timholy, line 2: If I'm *extending* a method I use `import`. If I'm just *calling* a method I use `using`, as a signal that there are no additional method of `countchildren` being defined here. But perhaps I'm missing a deeper reason for this change?

timholy:
Let's do the highlighting later? I'm a little worried about bulking this package up with too many dependencies, especially with https://github.com/JuliaDebug/Cthulhu.jl/pull/82 meaning that it might become a dependency of other packages.

We could also just use plain `printstyled`, no?

aminya:
Yes. Let me apply some quick `printstyled`. The `Highlights` seems like a lot of work!

examples/invalidations_blog.jl
aminya: Oh sorry. That is propagated from my closed #90 PR

aminya:
Now it looks easier for me to look at the result 😄

PR #93, opened 2020-06-15
Rename kw argument `blacklist` -> `exclusions`

timholy:
Long ago I chose the common term `blacklist` as a list of names to exclude from precompilation. Today I happened to notice that if you pause to think about it, this word evokes some ugly connotations (I'm not speaking historically here because I don't know the history, just what it evokes in my mind). Once noticed, I find it's not easy to un-notice it. I can't think of any good reason to have warm fuzzy feelings about this peculiarity of English, so I propose we change it. Just as "firefighter" is a far clearer description of a profession than "fireman," maybe we have a chance to eliminate a source of irritation while also increasing clarity.

@aminya, hope this is OK with you. Tests pass locally for me (and we'll soon find out on CI), but let me know if you see any ways this might break the bot that might not be covered by tests.

### DIFF
--- a/src/bot.jl
+++ b/src/bot.jl
@@ -15,14 +15,14 @@ end
 const UStrings = Union{AbstractString,Regex,AbstractChar}
 
 """
-    BotConfig(package_name::AbstractString ; blacklist, os, else_os, version, else_version, package_path, precompiles_rootpath, subst, tmin)
+    BotConfig(package_name::AbstractString ; exclusions, os, else_os, version, else_version, package_path, precompiles_rootpath, subst, tmin)
 
 Construct a SnoopCompile bot configuration. `package_name` is the name of the package. This object is passed to [`snoop_bot`](@ref)
 and [`snoop_bench`](@ref).
 
 You may supply the following optional **keyword** arguments:
 
-- `blacklist` : A vector of of Strings (or RegExp) to remove some precompile statements
+- `exclusions` : A vector of of Strings (or RegExp) to exclude some functions from being precompiled
 
 - `os`: A vector of of Strings (or RegExp) to support with precompile statements.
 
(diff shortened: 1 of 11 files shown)


src/bot.jl
```diff
@@ -128,9 +128,15 @@ function BotConfig(
     precompiles_rootpath::AbstractString = "$(dirname(dirname(package_path)))/deps/SnoopCompile/precompile",
     subst::AbstractVector = Vector{Pair{UStrings, UStrings}}(),
     tmin::AbstractFloat = 0.0,
-    yml_path::Union{String, Nothing} = nothing
+    yml_path::Union{String, Nothing} = nothing,
+    blacklist = nothing       # deprecated keyword
     )
 
+    if blacklist !== nothing
+        Base.depwarn("`blacklist` is deprecated, please use `exclusions` to pass a list of excluded names")
```
aminya, line 136: I think this will error 
```suggestion
        Base.depwarn(:blacklist, "`blacklist` is deprecated, please use `exclusions` to pass a list of excluded names")
```

test/bot/TestPackage2.jl/src/TestPackage2.jl
```diff
@@ -9,4 +9,5 @@ else
 
 end
 
+include("precompile_includer.jl")
```
aminya, line 12:
```suggestion

```

test/bot/TestPackage3.jl/src/TestPackage3.jl
```diff
@@ -12,4 +12,5 @@ else
 
 end
 
+include("precompile_includer.jl")
```
aminya, line 15:
```suggestion

```

test/bot/TestPackage4.jl/src/TestPackage4.jl
```diff
@@ -3,4 +3,5 @@ export hello4, domath4
 
 hello4(who::String) = "Hello, $who"
 domath4(x::Number) = x + 5
+include("precompile_includer.jl")
```
aminya, line 6:
```suggestion

```

test/bot/TestPackage1.jl/src/TestPackage1.jl
```diff
@@ -4,4 +4,5 @@ export hello, domath
 hello(who::String) = "Hello, $who"
 domath(x::Number) = x + 5
 
+include("precompile_includer.jl")
```
aminya, line 7:
```suggestion

```

PR #98, opened 2020-06-16
SnoopCompile split

### LINKED ISSUE #95: SnoopCompile may be too big
I'm a bit concerned about SnoopCompile's growing size. The main issue is that SnoopCompile has tools to spy on the system, but as it adds more of its own methods it introduces ever-more perturbations of that system.

Here's an interesting example: start a fresh REPL (with `--startup-file=no`) and cut-and-paste a "private" version of `@snoopr`, and then use it to snoop on `using SnoopCompile`:

```julia
julia> macro mysnoopr(expr)
           quote
               local list = ccall(:jl_debug_method_invalidation, Any, (Cint,), 1)
               Expr(:tryfinally,
                   $(esc(expr)),
                   ccall(:jl_debug_method_invalidation, Any, (Cint,), 0)
               )
               list
           end
       end
@mysnoopr (macro with 1 method)

julia> invs = @mysnoopr using SnoopCompile
3212-element Array{Any,1}:
  MethodInstance for isempty(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}})
  "invalidate_mt_cache"
  MethodInstance for ==(::Symbol, ::Int64)
  "invalidate_mt_cache"
  MethodInstance for ==(::Char, ::String)
  "invalidate_mt_cache"
  MethodInstance for ==(::Module, ::Module)
  "invalidate_mt_cache"
  MethodInstance for ==(::Expr, ::Int64)
  "invalidate_mt_cache"
  MethodInstance for ==(::TextDisplay, ::REPL.REPLDisplay{REPL.LineEditREPL})
  "invalidate_mt_cache"
  MethodInstance for ==(::Distributed.WorkerState, ::Distributed.WorkerState)
  "invalidate_mt_cache"
  MethodInstance for ==(::Array{String,1}, ::Bool)
  "invalidate_mt_cache"
  MethodInstance for join(::Base.Iterators.Take{Base.Iterators.Repeated{DataType}}, ::String)
  "invalidate_mt_cache"
  MethodInstance for join(::Tuple{Symbol}, ::String)
  "invalidate_mt_cache"
  MethodInstance for join(::Tuple{Symbol,Symbol}, ::String)
  "invalidate_mt_cache"
  MethodInstance for join(::Array{String,1}, ::String, ::String)
  "invalidate_mt_cache"
 ⋮
  MethodInstance for #_print#9(::Int64, ::Bool, ::Bool, ::Any, ::typeof(Pkg.TOML._print), ::Base.GenericIOBuffer{Array{UInt8,1}}, ::AbstractDict, ::Array{String,1})
 0
  MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:indent, :first_block, :sorted, :by),_A} where _A<:Tuple, ::typeof(Pkg.TOML._print), ::Base.GenericIOBuffer{Array{UInt8,1}}, ::AbstractDict, ::Array{String,1})
 1
  MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:indent, :sorted, :by),_A} where _A<:Tuple, ::typeof(Pkg.TOML._print), ::Base.GenericIOBuffer{Array{UInt8,1}}, ::AbstractDict, ::Array{String,1})
 1
  Tuple{Base.var"#sort!##kw",NamedTuple{(:by,),_A} where _A<:Tuple,typeof(sort!),Any}
  MethodInstance for #_print#9(::Int64, ::Bool, ::Bool, ::Any, ::typeof(Pkg.TOML._print), ::IOStream, ::AbstractDict, ::Array{String,1})
 0
  MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:indent, :first_block, :sorted, :by),_A} where _A<:Tuple, ::typeof(Pkg.TOML._print), ::IOStream, ::AbstractDict, ::Array{String,1})
 1
  MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:indent, :sorted, :by),_A} where _A<:Tuple, ::typeof(Pkg.TOML._print), ::IOStream, ::AbstractDict, ::Array{String,1})
 1
  Tuple{Base.var"#sort!##kw",NamedTuple{(:by,),_A} where _A<:Tuple,typeof(sort!),Any}
  (::Base.var"#sort!##kw")(::Any, ::typeof(sort!), d::OrderedCollections.OrderedDict) in OrderedCollections at /home/tim/.julia/packages/OrderedCollections/P6ntV/src/dict_sorting.jl:4
  "jl_method_table_insert"
  MethodInstance for fzero(::Base.Broadcast.Broadcasted{LinearAlgebra.StructuredMatrixStyle{T},Axes,F,Args} where Args<:Tuple where F where Axes where T)
 1
  MethodInstance for fzero(::Base.Broadcast.Broadcasted{LinearAlgebra.StructuredMatrixStyle{T},Nothing,typeof(-),_B} where _B<:Tuple where T)
 1
  MethodInstance for any(::typeof(ismissing), ::Any)
  "jl_method_table_insert"
  any(f, node::SnoopCompile.InstanceNode) in SnoopCompile at /home/tim/.julia/dev/SnoopCompile/src/invalidations.jl:49
  "jl_method_table_insert"

julia> length(unique(filter(x->isa(x, Core.MethodInstance), invs)))
1097
```
So, loading SnoopCompile itself invalidates more than 1000 MethodInstances. This is quite bad for a tool that is used to detect invalidations. By calling `invalidation_trees(invs)`, you can see that the overwhelming source of these invalidations is FilePaths.jl, with a small handful contributed by OrderedCollections. But as usual, the fix might be in Julia rather than FilePaths---without looking at FilePaths more carefully, this should not be interpreted as saying that FilePaths is necessarily doing something wrong.

One might worry about some of the other tools, too. SnoopCompile does not have a huge `__init__` function, so this won't force much inference/compilation, and what is done is really essential for `@snoopi` to be able to function. But there's a bit of top-level work in `botutils.jl` (it will compile methods for `Dict{String,String}`, if they aren't already), `deprecations.jl` (the `@eval` statements), `invalidations.jl` (surrounding `which` for `dummyinstance`), `parcel_snoopi.jl` (stuff around `lookup_kwbody_ex` and `default_exclusions`), and the regexs in `src/SnoopCompile.jl` itself. (There are a couple of others but I am pretty sure they aren't a problem or are strictly necessary for even the most minimal functionality.)

I am not quite sure about the best way to address this. One option would be to split out the absolute minimal "core" functionality into one or more packages of their own, and re-export it from SnoopCompile. People who care about this issue could use the micro-package(s) directly. The ability to put multiple packages into a single repo might be nice for this kind of thing.

aminya:
This initiates splitting SnoopCompile based on #95 

- [x] To start, I split the files into three different modules. 
- [ ] splitting the functions if needed 
	- [x] [`invalidations.jl`](https://github.com/aminya/SnoopCompile.jl/blob/SnoopCompile_Split/src/SnoopCompileCore/invalidations.jl): separate analysis from core. You know this part more @timholy. I think all of the functions except `@snoopr` should move to SnoopCompileAnalysis.
	- [ ] Can we separate the sorting of data i.e. [`sort_timed_inf`](https://github.com/aminya/SnoopCompile.jl/blob/93df49a20fa747da06a70ae613aab037663d457b/src/SnoopCompileCore/snoopi.jl#L41-L47) for `@snoopi`?
    - [ ] Why [`__inf_timing__`](https://github.com/aminya/SnoopCompile.jl/blob/SnoopCompile_Split/src/SnoopCompileCore/snoopi.jl#L3) is a global?
- [x] Determining the exports of each module, and import them from the higher-level module
- [x] Adjusting the code to reflect the separation. 
	- [x] SnoopCompileBot will need to [use external processes for loading SnoopCompileCore and then SnoopCompleAnalysis](https://github.com/timholy/SnoopCompile.jl/issues/95#issuecomment-644906289)
- [x] Separating Project.toml files for each module 
- [x] Splitting the tests and their extras in Project.toml
- [x] Splitting  CI 
- [ ] Splitting the docs
- [x] Deciding about the package registration:
	- [x] Should we try [multi-package one repo](https://github.com/JuliaLang/Pkg.jl/issues/1251)?
    - [ ]  or we should make these separate repositories?
- [ ] Use DocumenterTools to generate keys and secrets needed for GitHub Actions document building

To fix #95 

@timholy I gave you direct access to my repository.


timholy:
I started making some changes but things are moving around too much for me to submit a PR---it will be harder to rebase than to start from scratch. The main thing I am considering is to create `SnoopCompileInvalidations` as well. The analysis done there is very different from the more similar `@snoopi` and `@snoopc`, and moreover for https://github.com/JuliaDebug/Cthulhu.jl/pull/82 it might be nice to have this part be tiny. I don't insist on separating it from SnoopCompileAnalysis (maybe it would make sense to figure out what the load time of that "package," including `invalidations.jl`, really is, before deciding we have to make it even smaller).

aminya:
@timholy That is why I gave you access to my repository. I recommend that you make the changes in this branch.

We can separate it. That is not an issue. 🤔 I separated `snoopr` from the rest. We can make that a separate package. Does that affect the correctness of the result in Cthulhu?

aminya:
Something I noticed, is that we need to change some of our tests. For example, most of the `snoopi` tests should move to `SnoopCompileAnalysis`. I wrote the bot tests almost independently, and that was easier to split.

timholy:
I know I can submit a PR, but in the few minutes from me making changes you've made many that move files to new directories, etc, and the rebase isn't worth it. If you think things have settled down a bit then I can start to make my own tweaks.

aminya:
Can't you just push to this branch?

timholy:
Yes, I'll do that...but don't move code around drastically for a bit.

aminya:
Sure. For now, I will not do that.

timholy:
OK, I think this is mostly done. You'll have to do something like

```julia
(@v1.6) pkg> activate .
 Activating environment at `~/.julia/dev/SnoopCompile/Project.toml`

(SnoopCompile) pkg> dev SnoopCompileCore SnoopCompileAnalysis SnoopCompileBot

(SnoopCompile) pkg> activate
 Activating environment at `~/.julia/environments/v1.6/Project.toml`

(@v1.6) pkg> resolve
```

The bot tests seem to have partially moved into its subdirectory, so those tests are broken. But the rest seem to work.

timholy:
I'm not convinced we need to separate out SnoopCompileInvalidation, since the load time is so short. Let's see what the folks at Cthulhu say.

timholy:
I'm going to switch to other things, so feel free to make drastic changes again!

aminya:
For some time, I need to rewrite the bot package to [use external processes for loading SnoopCompileCore and then SnoopCompleAnalysis](https://github.com/timholy/SnoopCompile.jl/issues/95#issuecomment-644906289). So my changes will be mostly in that folder.

If we want to have three packages in one repository, shouldn't we separate the tests too?

timholy:
I don't think it's necessary, and in fact this way we make sure that old code will continue to work unmodified. The more we change the tests the more we risk accidental breakage. The only test I had to change was one that explicitly looked for a docstring in a particular module. Everything else now works out-of-the-box.

.github/workflows/ci.yml
```diff
@@ -1,7 +1,7 @@
 name: CI
-on:
-  - push
-  - pull_request
+# on:
+#   - push
+#   - pull_request
```
aminya, line 4: Feel free to enable the ci by uncommenting these
```suggestion
on:
  - push
  - pull_request
```

timholy:
Basically the split serves the purpose of separating the code that defines the module, but that's irrelevant for the tests. Since we'd submit PRs to the repo, it's fine if the tests live at that level.

If the tests take a while (which the `@snoopc` ones do), then someday there might be advantages in separating them out to make them easier to run independently, but for this stage I'm more worried about accidental breakage than speed of running tests.

aminya:
I was thinking that there is no way to do this without going to SnoopCompile 2.0. Now it seems that we might not need to do that!

aminya:
@timholy 
In the Bot tests, I get this error that ` @snoopi not defined`. I think that is because the parser tries to expand the macro before `using SnoopCompileCore` 🤔 (that is why we deprecated `@snoop_bot` macros).

SnoopCompileBot/Project.toml
```diff
@@ -0,0 +1,31 @@
+name = "SnoopCompileBot"
+uuid = "1d5e0e55-7d74-4714-b8d8-efa80e938cf7"
+author = ["Tim Holy <tim.holy@gmail.com>"]
```
timholy, line 3: We should change this to you.

SnoopCompileBot/Project.toml
```diff
@@ -0,0 +1,31 @@
...
+YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
+SnoopCompileAnalysis = "9ea4277c-da97-4c3a-afb0-537c066769de"
+SnoopCompileCore = "e2b509da-e806-4183-be48-004708413034"
+
+[compat]
+FilePathsBase = "0.9"
+YAML = "0.4"
+julia = "1"
+
+[extras]
```
timholy, line 18: Does the bot need all these for its tests?

SnoopCompileBot/src/snoop_bench.jl
```diff
@@ -1,37 +1,29 @@
...
-        empty!(SnoopCompile.__inf_timing__)
-        SnoopCompile.start_timing()
-        try
-            $(string(snoop_script));
-        finally
-            SnoopCompile.stop_timing()
+        using SnoopCompileCore
+
+        data = @snoopi tmin=$tmin begin
+            $snoop_script
```
timholy, line 10: Much better this way, I should have noticed this earlier.

aminya:
> Where do we stand with this? Tests pass for me locally. If we're close we can turn off draft mode.
> 

Did you try the bot tests too? I could not get them working.

> No need to back anything out of this, but for future I'd like to keep reorganizations like this totally separate from other changes. Building the docs in GitHub Actions, for example, could have waited for another PR. But no big deal, it was pretty easy to see what was what.

Yes, I agree! After I disabled CI, Travis was still running and throwing errors which forced me to remove it altogether.

SnoopCompileBot/src/snoop_bench.jl
aminya: It was like this originally, but because of those macro issues, I wanted to use `_snoopi` instead, which was not working at the time, and so I switched to direct code. 

Now I wanted to try this, I get `@snoopi` not defined. Does this work for you?

SnoopCompileBot/Project.toml
aminya: Fixed in https://github.com/timholy/SnoopCompile.jl/pull/98/commits/0ebdb9648b2cd6d941178d71d1b0ecef6c94526b

SnoopCompileBot/Project.toml
aminya: Thank you! https://github.com/timholy/SnoopCompile.jl/pull/98/commits/99001babd8e4e78696e22fb47c58a16f77b4299d

SnoopCompileBot/src/snoop_bot.jl
```diff
@@ -1,30 +1,46 @@
...
 function _snoopc_bot(snoop_script)
     return quote
+        using SnoopCompileCore
+
         @snoopc "compiles.log" begin
             $snoop_script
         end
-        data = SnoopCompile.read("compiles.log")[2]
+
+        using SnoopCompileAnalysis: read
```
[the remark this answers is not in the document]
aminya: But we need this in the line below to call `read`

timholy:
It's probably a top-level issue. Observe:
```julia
tim@diva:~/.julia/dev/FilePathsBase$ julia-master --startup-file=no -q
julia> eval(quote
           using SnoopCompileCore
           @snoopi sum([Float16(0.0)])
       end)
ERROR: LoadError: UndefVarError: @snoopi not defined
Stacktrace:
 [1] top-level scope
 [2] eval at ./boot.jl:331 [inlined]
 [3] eval(::Expr) at ./client.jl:446
 [4] top-level scope at REPL[1]:1
in expression starting at REPL[1]:3

julia> 
tim@diva:~/.julia/dev/FilePathsBase$ julia-master --startup-file=no -q
julia> ex = Expr(:toplevel, :(using SnoopCompileCore), :(@snoopi sum([Float16(0.0)])))
:($(Expr(:toplevel, :(using SnoopCompileCore), :(#= REPL[1]:1 =# @snoopi sum([Float16(0.0)])))))

julia> eval(ex)
2-element Array{Tuple{Float64,Core.MethodInstance},1}:
 (0.006245136260986328, MethodInstance for vect(::Float16))
 (0.00993800163269043, MethodInstance for sum(::Array{Float16,1}))
```

SnoopCompileBot/src/snoop_bot.jl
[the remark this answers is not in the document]
timholy: You scope it by module already, right?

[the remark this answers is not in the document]
aminya: I am not sure what do you mean. But I am lazy loading SnoopCompileAnalysis so it does not affect the snooping process.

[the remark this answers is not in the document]
timholy: If it loads `read` into `Main`, note that `Base.read != SnoopCompile.read`. So you could break `read` calls.

When you call `SomeModule.f(x)` you don't have to first say `using SomeModule: f`. You just need `using SomeModule`, and the scoping takes care of the rest.

aminya:
This seems to work:
```julia
tmin = 0.0
snoop_script = :(x=1)
ex = Expr(:toplevel, :(using SnoopCompileCore), quote
    data = @snoopi tmin = $tmin begin
        $snoop_script
    end
end)

```
However, I can't convert an expression to Cmd:
```julia
cmd = `$ex`
```
The following generates an extra `$` which throws errors.
```julia
'$(Expr(:toplevel, :(using SnoopCompileCore), quote
    #= untitled-7ec6d4b5f87a652319a23dc3aefd2636:4 =#
    data = #= untitled-7ec6d4b5f87a652319a23dc3aefd2636:4 =# @snoopi(tmin = 0.0, begin
                #= untitled-7ec6d4b5f87a652319a23dc3aefd2636:5 =#
                x = 1
            end)
end))'`
```

timholy:
An extra `eval` won't help. You're just generating a string, right? You can manage that manually. For example, maybe just strip the `begin`/`end` from your original version?

aminya:
I finally fixed the errors. I was using a Cmd to interpolate instead of String, which did not work.

aminya:
@timholy I think either we need to register the individual packages before we can make the CI fully working.

CI now passes for Core and Analysis because they are using top-level SnoopCompile. But SnoopCompileBot tests fail because Pkg is not able to instantiate the SnoopCompileBot environment.

timholy:
Yes, I think we will have to register it. Any reason not to merge this? Once merged I will submit the registration requests. We will have to wait 3 days, during which time `master` is effectively broken, though the fix (to `dev` each sub-package) is very simple.

timholy:
Interesting question: will this really solve the "co-developed" problem? Or if you're submitting a PR to SnoopCompile, will it run the tests with an older version of one of the submodules? Interesting questions...

.github/workflows/ci.yml
```diff
@@ -11,9 +11,9 @@ jobs:
       fail-fast: true
       matrix:
         version:
-          - '1.0.5'
+          - '1.0'
```
timholy, line 14: If we actually need to be this specific, let me know.

timholy:
Hmm, curious why the tests weren't triggered. Did I bork the `ci.yml` file?

timholy:
This almost makes it through, fails at [this line](https://github.com/aminya/SnoopCompile.jl/blob/85623453d5313f8f8a3bfff514062b71a5c2e64c/SnoopCompileBot/test/runtests.jl#L433).

aminya:
The CI tests now pass. 🚀

.github/workflows/ci.yml
aminya: Not anymore. Initially, I was using very accurate versions. But later I decided to assume that all the patch versions will have one precompile folder and the user will always update to the latest patch.

aminya:
Documenter passes as well. You will just need to add the keys:

```julia
julia> using DocumenterTools, SnoopCompile
julia> DocumenterTools.genkeys(SnoopCompile)
```

timholy:
https://github.com/JuliaRegistries/Registrator.jl/issues/298

timholy:
Documenter keys are set up. Once the registrations are done we can submit the doc changes.

timholy:
https://github.com/JuliaRegistries/General/pull/16663

KristofferC:
FWIW, I think that the subpackages should have their own tests. The whole reason for splitting them into separate packages is presumably that they have their own API and can be used/tested individually?

src/SnoopCompile.jl
```diff
@@ -1,28 +1,31 @@
...
-export timesum  # @snoopi and @snoopc are exported from their files of definition
+using SnoopCompileAnalysis
 
-const anonrex = r"#{1,2}\d+#{1,2}\d+"         # detect anonymous functions
-const kwrex = r"^#kw##(.*)$|^#([^#]*)##kw$"   # detect keyword-supplying functions
-const kwbodyrex = r"^##(\w[^#]*)#\d+"         # detect keyword body methods
-const genrex = r"^##s\d+#\d+$"                # detect generators for @generated functions
-const innerrex = r"^#[^#]+#\d+"               # detect inner functions
+# needed for snoopc
+using SnoopCompileAnalysis: parcel, read, write, parse_call, format_userimg
```
KristofferC, line 9: This doesn't make sense to me. These functions are not used anywhere, how can they be "needed" for snoopc?

src/SnoopCompile.jl
```diff
@@ -1,28 +1,31 @@
...
-    include("invalidations.jl")
+using SnoopCompileBot
+export BotConfig, snoop_bot, snoop_bench
+export timesum, pathof_noload, GoodPath
+if isdefined(SnoopCompileBot, Symbol("@snoopiBench"))
+    # deprecated names
+    export @snoopiBench, @snoopiBot, @snoopi_bench, @snoopi_bot
 end
 
+export SnoopCompileCore, SnoopCompileAnalysis, SnoopCompileBot
```
KristofferC, line 29: Why export these?

timholy:
> FWIW, I think that the subpackages should have their own tests. The whole reason for splitting them into separate packages is presumably that they have their own API and can be used/tested individually?

We should make it more fine-grained and then add "integration" tests, e.g., that what `@snoopr` returns is valid input for `invalidation_trees`. But this was easy and I am swamped by more urgent things. This loses nothing compared to where we were before. Just because we need more fine-grained control over code-loading does not impose a new burden on how fine-grained are tests need to be.

src/SnoopCompile.jl
timholy: Formerly we told people to call `SnoopCompile.parcel(data)`. This is compatibility glue.

src/SnoopCompile.jl
timholy: Indeed we might want to get rid of them. Will try.

timholy:
I'm switching to tilde bounds on the package versions, to more cleanly lock versions together. Doesn't affect the patch release but I think that should be fine in most cases.

PR #101, opened 2020-06-23
Add support for Cthulhu

timholy:
This is the final tool (so far) for fixing invalidations. `ascend` makes it much easier to identify a fruitful place to intervene and fix inference problems.

This also adds tools for bumping the versions across-the-board. What do you think of this, @aminya? Overall I think our best strategy is to keep all version numbers in lock-step, and that's the workflow made easy with these tools. When you want to bump the version, you say `bump_version(v"1.7.0")`. After the PR merges, update your local master branch to the latest, check out a branch of `registries/General`, and then `register_all()`. Then submit your `registries/General` as a PR.

PR #104, opened 2020-06-24
Fixing addtestdep not defined

aminya:
In this new configuration, people cannot call the internal functions of each package through SnoopCompile. 

We should probably use `@reexport` to solve this issue.

### DIFF
--- a/SnoopCompileBot/src/botutils.jl
+++ b/SnoopCompileBot/src/botutils.jl
@@ -236,7 +236,7 @@ function pathof_noload(package_name::String)
 end
 
 ################################################################
-
+export addtestdep
 import Pkg
 """
 Should be removed once Pkg allows adding test dependencies to the current environment
--- a/src/SnoopCompile.jl
+++ b/src/SnoopCompile.jl
@@ -20,7 +20,7 @@ end
 
 using SnoopCompileBot
 export BotConfig, snoop_bot, snoop_bench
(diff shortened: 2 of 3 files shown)


aminya:
This does not solve the issue, and I cannot run `SnoopCompile.addtestdep()`.
https://github.com/JuliaMusic/MusicXML.jl/runs/801791877#step:4:175

timholy:
You just to say `using SnoopCompileBot: privatefunction` in `src/SnoopCompile.jl`. I'll split what I have in #101 out and submit separately without bumping the version.

SnoopCompileBot/src/botutils.jl
```diff
@@ -236,7 +236,7 @@ function pathof_noload(package_name::String)
 end
 
 ################################################################
-
+export addtestdep
```
timholy, line 239: Don't do this. Use `using SnoopCompileBot: addtestdep` instead.

src/SnoopCompile.jl
```diff
@@ -20,7 +20,7 @@ end
 
 using SnoopCompileBot
 export BotConfig, snoop_bot, snoop_bench
-export timesum, pathof_noload, GoodPath
+export timesum, pathof_noload, GoodPath, addtestdep
```
timholy, line 23: No, definitely not!

aminya:
#107 solves this

Issue #105, opened 2020-06-24
SnoopCompile 1.6.0 is broken


aminya:
We registered the inner packages, but we did not find time to check if the master will work when the inner packages are registered. We had [a hack script that solved this issue](https://github.com/timholy/SnoopCompile.jl/pull/98/commits/bb56316297de70d401e84433bf84064cd9cee9fe), but by removing it now we have a broken master that is tagged as 1.6. 

I think we should have registered the inner packages first, and then registered the top-level package separately.

Also, I am getting the error in #104, which we were not getting it before the split.

KristofferC:
Definitely don't do something like https://github.com/timholy/SnoopCompile.jl/pull/98/commits/bb56316297de70d401e84433bf84064cd9cee9fe

timholy:
Closed by #107

aminya:
Now that 1.6.1 is registered, I still get errors when running the script online:
https://github.com/JuliaMusic/MusicXML.jl/runs/805777098

```
ERROR: LoadError: ArgumentError: Package SnoopCompileBot not found in current path:
- Run `import Pkg; Pkg.add("SnoopCompileBot")` to install the SnoopCompileBot package.
```

KristofferC:
> Apparently, Pkg.add("SnoopCompile") does not add the subdirs. 

This doesn't make any sense. It adds the package `SnoopCompile`.

> This is a breaking change for the bot API if we have to wait for Pkg to fix this in some later Julia version.

Just to set expectations, there is nothing to fix in Pkg here. To try to explain this more, there is nothing more special about the relation SnoopCompile <-> SnoopCompileBot than it is to SnoopCompileBot <-> YAML. And just because you add SnoopCompileBot, it doesn't mean that YAML is somehow available to use. If you want to use YAML you add it, if you want to use SnoopCompileBot you add it.

Also, I want to reiterate that https://github.com/aminya/SnoopCompile.jl/blob/depsbuild/deps/build.jl is completely wrong. The build script runs in a separate temporary environment and will not have an effect on the current one. Calling `Pkg` functions in the build script is fundamentally not what you want to do.

aminya:
Thanks for the answers. I want all the sub-packages to be available under the global environment. For backward compatibility, I need to add a script to add those sub-packages inside SnoopCompile.

I think something like this will solve the problem:
```julia
[...]

PR #107, opened 2020-06-24
Various fixes (mostly for the split)

timholy:
This fixes a number of issues, most of which are leftovers from splitting the package and most of which only concern the docs. The `reason` change fixes a bug when invalidations stem from both method insertions and method deletions, and the Manifest.toml file is a workaround for https://github.com/JuliaLang/Pkg.jl/issues/1874. We'll have to remember to update the Manifest before every PR, presumably, and I'm not sure it will be robust across Julia versions, but I think it's the best we can do for now.

@aminya, if #105 is still an issue even after this, you may want to add some additional `using SnoopCompileBot: some_private_function` statements to `SnoopCompile`, see the existing models. I'll wait for your comments on this before merging (feel free to submit a PR against this branch).


.github/workflows/ci.yml
```diff
@@ -27,6 +27,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
+      - run: julia -e 'using Pkg; pkg"dev SnoopCompile"; pkg"dev SnoopCompileCore SnoopCompileAnalysis SnoopCompileBot"'
```
KristofferC, line 30: Maybe

```
julia --project -e 'using Pkg; Pkg.develop([PackageSpec(path="SnoopCompileCore"),
                                            PackageSpec(path="SnoopCompileAnalysis"),
                                            PackageSpec(path="SnoopCompileBot")])'

```

.github/workflows/ci.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
+      - run: julia --project -e 'using Pkg; Pkg.develop([PackageSpec(path="SnoopCompileCore"),
```
KristofferC, line 30: I wonder if this needs some escaping on Windows?

```

Run julia --project -e 'using Pkg; Pkg.develop([PackageSpec(path="SnoopCompileCore"), PackageSpec(path="SnoopCompileAnalysis"), PackageSpec(path="SnoopCompileBot")])'
  julia --project -e 'using Pkg; Pkg.develop([PackageSpec(path="SnoopCompileCore"), PackageSpec(path="SnoopCompileAnalysis"), PackageSpec(path="SnoopCompileBot")])'
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
ERROR: UndefVarError: SnoopCompileCore not defined
Stacktrace:
 [1] top-level scope at none:0
```

pfitzseb:
```
julia --project -e 'using Pkg; Pkg.develop([PackageSpec(path=\"SnoopCompileCore\"),
                                            PackageSpec(path=\"SnoopCompileAnalysis\"),
                                            PackageSpec(path=\"SnoopCompileBot\")])'
```
works fine for me on Windows.

aminya: This is becayse PowerShell needs `\` for scaping `"`. Instead, we can add the following before `jobs`. It will use bash for all operating systems:
```
defaults:
  run:
    shell: bash
```

aminya, line 30: Is this the correct thing to do? @timholy When someone adds `SnoopCompile`, these packages are not developed by default.

timholy: I'm not even sure how to interpret that error message. Anyone have any ideas?

timholy: When they add it, it will be registered. In principle, this is a good workaround for https://github.com/JuliaLang/Pkg.jl/issues/1874 (better than committing the Manifest).

aminya: @timholy Let's first choose bash as I suggested. Then we can see if it errors again.

aminya: Because this is similar to https://github.com/timholy/SnoopCompile.jl/commit/bb56316297de70d401e84433bf84064cd9cee9fe, I am confused. If developing is a bad thing we should not do it at all (including in CI). If it is a good thing when we don't do it by default (when someone adds the package).

johnnychen94: > If running this script is a bad thing we should not do it at all (including in CI).

For test usage, it'll need a `dev`ed version of these sub-packages to make sure we can submit three releases altogether. While on the other hand, common users should just install the released version of these subpackages as dependencies.

`Pkg.develop` in `deps/build.jl` is like pulling arbitrary code from the internet and install it; there's no quality and version control here.

aminya: That means people cannot develop SnoopCompile directly, and we might want to add this to the contribution documentation.

timholy: Oh, looks like I just needed to refresh my browser.

aminya:
I think we should restart the failed jobs. That was probably a [connection issue](https://github.com/julia-actions/setup-julia/issues/35).

timholy:
Looks like doing the `dev` from within the `ci.yml` script works well. Thanks for all the suggestions, everyone! I'll merge this soon if there are no objections. @aminya, be sure to test Bot functionality and add any more `using` entries you need.

timholy:
Instead of restarting, any additions you want? Might as well test those as the same time.

aminya:
I am testing this branch here: https://github.com/JuliaMusic/MusicXML.jl/actions?query=workflow%3ASnoopCompile

aminya:
This fails.
https://github.com/JuliaMusic/MusicXML.jl/runs/804800745?check_suite_focus=true#step:4:91
```julia
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Resolving package versions...
ERROR: Unsatisfiable requirements detected for package SnoopCompileBot [1d5e0e55]:
 SnoopCompileBot [1d5e0e55] log:
 ├─possible versions are: 1.6.0 or uninstalled
 └─restricted to versions 1.6.1-1.6 by SnoopCompile [aa65fe97] — no versions left
   └─SnoopCompile [aa65fe97] log:
     ├─possible versions are: 1.6.1 or uninstalled
     └─SnoopCompile [aa65fe97] is fixed to version 1.6.1
```

timholy:
Packages shouldn't `dev` SnoopCompile. I'm asking you to run it locally on your machine and see if all the names are in their proper places now with no changes to the old snoopbot scripts. Once you tell me that, we can merge & release and then MusicXML.jl will have no problem getting a consistent 1.6.1 release.

src/SnoopCompile.jl
```diff
@@ -25,7 +25,6 @@ if isdefined(SnoopCompileBot, Symbol("@snoopiBench"))
     # deprecated names
     export @snoopiBench, @snoopiBot, @snoopi_bench, @snoopi_bot
 end
-
-export SnoopCompileCore, SnoopCompileAnalysis, SnoopCompileBot
+using SnoopCompileBot: standardize_osname, JuliaVersionNumber
```
aminya, line 28:
```suggestion
using SnoopCompileBot: standardize_osname, JuliaVersionNumber, addtestdep
```

aminya: Does this make `addtestdep` under `SnoopCompile`?

timholy:
If that's it...I have a commit queued to add the `dev` strategy to Documenter.yml, but let's see how this does otherwise through CI. If you have nothing else, then I think this may be good to go.

aminya:
I tested this on MusicXML locally and it works now.

I made a dev script, which allows me to dev SnoopCompile for different environments quickly.
https://github.com/timholy/SnoopCompile.jl/pull/108

PR #108, opened 2020-06-24
Dev script

aminya:
This is a script for my personal use, so I can `dev` SnoopCompile easily.


timholy:
Can you just keep that script on your own hard drive?

aminya:
> Can you just keep that script on your own hard drive?

That is possible. We can also add a CONTRIBUTING.md and mention it there.

aminya:
I hope this dev gets baked into Pkg itself. 🙏

PR #110, opened 2020-06-24
Minor doc tweaking

aminya:


docs/src/snoopr.md
```diff
@@ -13,6 +13,14 @@ end
 ```
 
 Invalidations occur when there is a danger that new methods would supersede older methods in previously-compiled code.
+
+To record the invalidations caused by defining new methods, use the following syntax:
```
timholy, line 17: Since you're providing a little more explanation about what's going on...technically, `@snoopr` does the recording, `invalidation_trees` does the analysis.

docs/src/snoopr.md
```diff
@@ -128,7 +137,7 @@ In this example, there were four different methods that triggered invalidations,
 You can see that collectively more than a thousand independent compiled methods needed to be invalidated; indeed, the last
 entry alone invalidates 1027 method instances:
 
-```
+```julia
```
timholy, line 140: I thought this is the default for Documenter (not for GitHub), but now I can't find evidence of that.

docs/src/snoopr.md
```diff
@@ -259,7 +268,7 @@ Variables
   container::Array{AbstractFloat,1}
 
 Body::Int64
-    @ REPL[3]:1 within `callf'
+    @ REPL[3]:1 within callf
```
timholy, line 271: Prints with the backticks for me?

docs/src/snoopr.md
aminya: I guess it is better to stay on the safe side 😁

docs/src/snoopr.md
aminya: This backtick breaks the code blocks of the markdown. I thought removing it does not change the content much 🤔 
Can we escape ` ?

docs/src/snoopr.md
aminya: Does it look better now?
https://github.com/timholy/SnoopCompile.jl/pull/110/commits/82973b03520aafe9e6ad89876cea900a4ef7f692

docs/src/snoopr.md
```diff
@@ -13,6 +13,20 @@ end
...
 Invalidations occur when there is a danger that new methods would supersede older methods in previously-compiled code.
+
+To record the invalidations caused by defining new methods, use `@snoopr` from SnoopCompileCore:
+```julia
+using SnoopCompileCore
+invalidations = @snoopr begin
+ # new methods definition
+end
+```
+and use `invalidation_trees` fomr SnoopCompileAnalysis to make the `trees`:
```
timholy, line 24:
```suggestion
and use `invalidation_trees` from SnoopCompileAnalysis to aggregate the information as a collection of [tree structures](https://en.wikipedia.org/wiki/Tree_structure):
```

PR #112, opened 2020-06-27
Fix SnoopCompileBot

### LINKED ISSUE #111: ERROR: LoadError: ArgumentError: Package SnoopCompileBot not found in current path
I am trying to use SnoopCompileBot via Github Actions for Plots.jl and get the following error:
```julia

Run julia --project -e 'include("deps/SnoopCompile/snoop_bot.jl")'
[ Info: /home/runner/work/Plots.jl/Plots.jl/src/precompile_includer.jl file will be created/overwritten
[ Info: SnoopCompile will try to write  "include("precompile_includer.jl")" before end of the module in /home/runner/work/Plots.jl/Plots.jl/src/Plots.jl. Assume that the last `end` is the end of a module.
ERROR: LoadError: ArgumentError: Package SnoopCompileBot not found in current path:
- Run `import Pkg; Pkg.add("SnoopCompileBot")` to install the SnoopCompileBot package.

Stacktrace:
 [1] require(::Module, ::Symbol) at ./loading.jl:892
 [2] top-level scope at /home/runner/.julia/packages/SnoopCompileBot/K9naN/src/snoop_bot.jl:112
 [3] eval at ./boot.jl:331 [inlined]
 [4] snoop_bot(::BotConfig, ::String, ::Module; snoop_mode::Symbol) at /home/runner/.julia/packages/SnoopCompileBot/K9naN/src/snoop_bot.jl:182
 [5] snoop_bot at /home/runner/.julia/packages/SnoopCompileBot/K9naN/src/snoop_bot.jl:177 [inlined] (repeats 2 times)
 [6] top-level scope at /home/runner/work/Plots.jl/Plots.jl/deps/SnoopCompile/snoop_bot.jl:3
 [7] include(::String) at ./client.jl:439
 [8] top-level scope at none:1
in expression starting at /home/runner/work/Plots.jl/Plots.jl/deps/SnoopCompile/snoop_bot.jl:3
##[error]Process completed with exit code 1.
```
I think I followed the instructions in the [docs](https://timholy.github.io/SnoopCompile.jl/stable/bot/) but I am probably doing something stupid. Here's the test PR on my fork: https://github.com/daschw/Plots.jl/pull/2. I would really appreciate some hints to what I am doing wrong. Thanks!

aminya:
- [x] To fix #111 and #105 
- [x] Add workflow CI:
	I added workflow_CI, so we have a better idea of what is going on with SnoopCompileBot. This will prevent us from tagging broken versions from now on.
- [x] ⚠️  Should not be merged until we make sure that the online tests pass here: https://github.com/JuliaMusic/MusicXML.jl/actions?query=workflow%3ASnoopCompile

- [x] ~~This is just a draft now, and I will experiment with different methods here. So wait until a solution is found.~~

We can revert the changes of Project.toml and src/SnoopCompile.jl if you prefer that. Using the workflow_CI, we can see if the package breaks by doing that.


deps/build.jl
```diff
@@ -0,0 +1,27 @@
+rootdir = dirname(@__DIR__)
```
KristofferC, line 1: This PR should absolutely not be merged with this. I have no idea what it is trying to do but it is completely wrong and will mess up the projects for users (even in environments they don't have active right now!).

I have said over and over that doing these types of Pkg-shenanigans in the build script is incorrect and will not work properly. However, since I am clearly getting ignored, I'll just stop.

aminya: Dear @KristofferC, I do not like or want to do this, but I have any other solution. All of my tests simply fail. I am experimenting with brute-force solutions to see where the problem comes from.

aminya:
This finally passes the tests https://github.com/JuliaMusic/MusicXML.jl/actions/runs/149958419. This was a cumbersome experience. 😥 

This solution is very ugly :unamused:, but that is the only one I could find. Let me know if you know a better trick.

aminya:
Because of importing conflicts, this breaks the direct use of SnoopCompileBot in the snooping script with `using SnoopCompileBot` in favor of the old `using SnoopCompile`. 

Because of that, the tests here fail, but that does not matter much. Later in SnoopCompile 2, we can fix this. 

For now, we can change the tests to `using SnoopCompile`, and test the SnoopCompileBot as part of SnoopCompile.

aminya:
I added workflow_CI, so we have a better idea of what is going on with SnoopCompileBot. This will prevent us from tagging broken versions from now on.

timholy:
Superseded by #113

PR #113, opened 2020-06-28
Version 1.6.2

timholy:
Fixes the bot on 1.6. Tested against MusicXML v0.3.2. I had to delete the `exhausive` keywords from the snoop scripts; that was never in a released version of SnoopCompile, so that isn't cheating.

@aminya, one thing I had to do to make it work was execute the instructions in https://github.com/timholy/SnoopCompile.jl/issues/105#issuecomment-649345931 in the MusicXML environment, since the bot launches with `--project`. Interestingly, I had to manually edit the paths in the resulting `MusicXML/Manifest.toml` manually: doing the `rm SnoopCompileCore ...` deleted the `path=` statement from the Manifest.toml and reverted to the git tree sha.

### DIFF
--- a/SnoopCompileBot/src/snoop_bench.jl
+++ b/SnoopCompileBot/src/snoop_bench.jl
@@ -4,7 +4,7 @@ function _snoopi_bench_cmd(snoop_script)
     return quote
         global SnoopCompile_ENV = true
 
-        using SnoopCompileCore
+        using SnoopCompile
 
         data = @snoopi tmin=$tmin begin
             $snoop_script
@@ -12,7 +12,7 @@ function _snoopi_bench_cmd(snoop_script)
 
         global SnoopCompile_ENV = false
 
-        using SnoopCompileBot: timesum
+        using SnoopCompile: timesum
         @info( "\nTotal inference time (ms): \t" * string(timesum(data, :ms)))
     end
 end
@@ -56,7 +56,7 @@ function _snoop_bench(config::BotConfig, snoop_script::Expr, test_modul::Module 
     out = quote
         package_sym = Symbol($package_name)
         ################################################################
-        using SnoopCompileBot
+        using SnoopCompile
         @info("""------------------------
         Benchmark Started
         ------------------------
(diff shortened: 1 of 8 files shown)


SnoopCompileBot/src/snoop_bench.jl
```diff
@@ -4,15 +4,15 @@ function _snoopi_bench_cmd(snoop_script)
     return quote
         global SnoopCompile_ENV = true
 
-        using SnoopCompileCore
+        using SnoopCompile
```
aminya, line 7: This reverts the package to the pre-split era. I wanted to avoid doing this in #112. Now online usage of SnoopCompile will give the wrong results.

aminya:
> Fixes the bot on 1.6. Tested against MusicXML v0.3.2. I had to delete the `exhausive` keywords from the snoop scripts; that was never in a released version of SnoopCompile, so that isn't cheating.
> 

I had fixed the old deprecations on the master branch. But it is not registered it yet.  

> @aminya, one thing I had to do to make it work was execute the instructions in [#105 (comment)](https://github.com/timholy/SnoopCompile.jl/issues/105#issuecomment-649345931) in the MusicXML environment, since the bot launches with `--project`. Interestingly, I had to manually edit the paths in the resulting `MusicXML/Manifest.toml` manually: doing the `rm SnoopCompileCore ...` deleted the `path=` statement from the Manifest.toml and reverted to the git tree sha.

Registration of the package will probably solve this. I bypassed the registration headache by just including SnoopCompileBot directly in SnoopCompile.jl.  See its successful run:
 https://github.com/JuliaMusic/MusicXML.jl/actions/runs/150097985

SnoopCompileBot/src/snoop_bench.jl
timholy: Why? It's just a question of whether you load SnoopCompile + bot, or just the bot.

But there really isn't a choice: if you wanted it this way we needed to call that release 2.0. We can move there, but I was trying to maintain backwards compatibility.

aminya: Here we are loading the SnoopCompile internally, which loads all the packages. So it does not matter what the user wants to do. 

The problem of the bot breaking in version 1.6.0 and 1.6.1 is that Pkg cannot handle adding these to the global environment. That is why we need to **add them before using** like how I did in #112. There is no difference between SnoopCompile or SnoopCompileBot. That is why CI fails here.
https://github.com/timholy/SnoopCompile.jl/pull/113/checks?check_run_id=816357714#step:6:205

timholy:
Sure, I'd be happy to have the workflow file join this. Let's see if these tests pass and then we can merge that. I'll check in the morning.

PR #114, opened 2020-06-28
Workflow CI

aminya:
Adds workflow CI:
I added workflow_CI, so we have a better idea of what is going on with SnoopCompileBot. This will prevent us from tagging broken versions from now on.

cherry-picked from #112


timholy:
Now that 1.6.2 is registered, let's re-run CI here.

timholy:
I'm a bit confused by the error message, any idea?

aminya:
He

aminya:
@timholy Because you were the triggering actor of this action (opening and closing the PR), it thought that the branch is on your profile.

I have made a [pull request](https://github.com/julia-actions/GitHubActions.jl/pull/1) to allow more robust ways of finding the PackageSpec for a repository. That will solve the issue eventually.

aminya:
Most of the tests pass. Pkg fails to resolve SnoopCompileBot on Windows in one of the tests:
https://github.com/timholy/SnoopCompile.jl/pull/114/checks?check_run_id=819024363

Also, on Julia 1.0.5, subdir package adding seems to be broken altogether.
https://github.com/timholy/SnoopCompile.jl/pull/114/checks?check_run_id=819024207

timholy:
Interesting. This is only because this uses `SnoopCompileBot` directly rather than via the `SnoopCompile` wrapper, right?

The split is only really urgent for `@snoopr`. When we want to run `@snoopr` with a bot we will need to go to v2.0 and not support Julia 1.0, probably?

aminya:
> Interesting. This is only because this uses `SnoopCompileBot` directly rather than via the `SnoopCompile` wrapper, right?

We can test that by using more versions in the Presplit_Compatibility job.

aminya:
This is ready as well

PR #117, opened 2020-07-01
Try to get the docs deploying

timholy:
For some reason the docs haven't deployed since 1.5. [This run](https://github.com/timholy/SnoopCompile.jl/runs/824519353) suggests, if anything, that it's due to a lack of `push_preview` (though that seems weird). Let's try this, though, and see if it fixes it.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -11,5 +11,6 @@ makedocs(
 )
 
 deploydocs(
-    repo = "github.com/timholy/SnoopCompile.jl.git"
+    repo = "github.com/timholy/SnoopCompile.jl.git",
+    push_preview=true
 )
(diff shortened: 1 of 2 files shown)


aminya:
Did you add the proper keys?
```julia
julia> using DocumenterTools, SnoopCompile
julia> DocumenterTools.genkeys(SnoopCompile)
```

timholy:
Yes, you can see it found the keys from the logs:
```
│ - ✔ ENV["DOCUMENTER_KEY"] exists exists
```

timholy:
https://timholy.github.io/SnoopCompile.jl/previews/PR117

aminya:
What happens if you run the script offline manually?

timholy:
Works fine. (I always do that whenever working on the docs.)

timholy:
You get some warnings when you do so, we'll have that until https://github.com/JuliaDocs/Documenter.jl/pull/1352 merges and gets released.

aminya:
I think it is better to test this on the master instead. Merge it there and see if the docs get updated.

PR #118, opened 2020-07-01
Using sub packages in SnoopCompileBot

aminya:
Together with my next pull request which will add snoopr to SnoopCompileBot, this will should be tagged as 1.8

As you see, [despite adding the packages in advance](https://github.com/aminya/SnoopCompile.jl/blob/BotSubPackage/.github/workflows/workflow_ci.yml#L49), the sub-packages are not found in the `test_modul` and my compatibility function is triggered to fix that.
https://github.com/timholy/SnoopCompile.jl/pull/118/checks?check_run_id=825156566#step:5:6

### DIFF
--- a/.github/workflows/workflow_ci.yml
+++ b/.github/workflows/workflow_ci.yml
@@ -15,21 +15,15 @@ jobs:
         test_package:
           - ./SnoopCompileBot/test/TestPackage1.jl
     steps:
-      - run: |
-          if [ -z "${{github.head_ref}}" ]; then
-            echo ::set-env name=SnoopCompilePackageSpec::$(echo "PackageSpec(url = \"https://github.com/${{github.actor}}/SnoopCompile.jl\", rev = \"${GITHUB_REF#refs/heads/}\")")
-          else
-            echo ::set-env name=SnoopCompilePackageSpec::$(echo "PackageSpec(url = \"https://github.com/${{github.actor}}/SnoopCompile.jl\", rev = \"${{ github.head_ref }}\")")
-          fi
       - uses: actions/checkout@v2
-      - name: Add SnoopCompileBot
+      - name: Add SnoopCompileBot and friends
         run: |
-          julia -e 'using Pkg; Pkg.develop( PackageSpec(path="SnoopCompileBot") )'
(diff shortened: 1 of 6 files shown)


timholy:
I think perhaps we should just jump to 2.0 and not try to keep getting access to the subpackages via indirect means. In 2.0 we can make it a requirement that the user add SnoopCompileBot directly.

aminya:
Going to 2.0 does not help. Even with adding the sub-packages in advance, my script is still triggered! Either I am missing something or Pkg is not doing something right.

> As you see, [despite adding the packages in advance](https://github.com/aminya/SnoopCompile.jl/blob/BotSubPackage/.github/workflows/workflow_ci.yml#L49), the sub-packages are not found in the `test_modul` and my compatibility function is triggered to fix that.
https://github.com/timholy/SnoopCompile.jl/pull/118/checks?check_run_id=825156566#step:5:6

timholy:
There are still failures here, and there have been many since #114. What if we just stop trying to address SnoopCompileBot itself and only use SnoopCompile?

I have to confess I'm getting a bit frustrated. It might be time to consider moving the bot to a separate repository/package.

aminya:
> There are still failures here, and there have been many since #114. What if we just stop trying to address SnoopCompileBot itself and only use SnoopCompile?
> 

I suppressed the jobs that fail in #121. These were failing from the beginning. One is for Julia 1.0.5 which misses some methods, and the other one is on Windows, which Pkg fails running `resolve`.

> I have to confess I'm getting a bit frustrated. It might be time to consider moving the bot to a separate repository/package.

We have had a hard time after splitting. Maybe the mono-repo feature of Pkg was not still ready. However, our experience helped it to reveal its bugs. I will create separate issues for the things we see here in Pkg. Hopefully, they will get some attention.

I am OK with moving things to another repository. SnoopCompile.jl is backward compatible, and nothing changes if we move SnoopCompileBot to another repo.

timholy:
I think when we split, it should be a clean split. Since the bot was part of previous SnoopCompile releases, removing it will be a breaking change. People who want the bot should install it directly, rather than via SnoopCompile.

If necessary, that will free SnoopCompile to going back to being a mono-repo, since almost all the invalidations were cause by the bot's dependencies.

PR #119, opened 2020-07-01
Fix TagBot so it can produce docs for releases

timholy:
Found in https://github.com/JuliaDocs/Documenter.jl/issues/1277#issuecomment-612701975


timholy:
Hmm, did I break your tests with #116? I found that they have to be run from the top-level directory, hence the modification. But I don't see how it's related since it doesn't seem to be used by the workflow script.

timholy:
No dice, https://github.com/timholy/SnoopCompile.jl/runs/826229844

Can we go back to travis for the docs?

aminya:
Sure if that helps! Having docs is more important than anything else.

Issue #99, opened 2020-06-17
SnoopCompileBot deprecations notice


aminya:
Some packages are using `.github/workflows/SnoopCompile.yml`, and inside that, they `add SnoopCompile` and use it. Is there a way that CompatHelper allows us to find these packages and give them some notification that SnoopCompile is updated and they might need to update some of their code?

[Deprecation warnings are just shown passed in GitHub actions and the users may not notice anything](https://github.com/JuliaLang/julia/issues/36326).

One solution might be using CompatHelper:
https://github.com/bcbi/CompatHelper.jl/issues/218

timholy:
I guess the risk is they might not check the logs? I think it's on them :smile:.

aminya:
Part of it is that not very often people read CI log (unless it errors), and that is why we need to have a better way to inform them. My Julia issue was closed, but @christopher-dG mentioned that he can make a new package, that gives us the ability to push the warnings to GitHub's warning system.

christopher-dG:
The actual code is basically done, just needs some grunt work https://github.com/julia-actions/GitHubActions.jl

aminya:
@christopher-dG That is awesome. Could we rename this package to CILogger instead, and implement a different style of logging in one package? We can add more services like Travis, Azure, etc.

christopher-dG:
No, this one is going to be just GitHub Actions (it does more than logging). You could make a meta-package of sorts that aggregates various similar packages if you like.

PR #122, opened 2020-07-02
missing includer issue

aminya:
To fix missing includer issue in https://github.com/JuliaPlots/Plots.jl/issues/2838

⚠️ emergency patch 

- Adds gitattributes to prevent line ending changing during uploads from different operating systems
- removes previous bash script that did a similar thing
- throws an error for anyone who uses the bash script (there is no other way to let them know 😞)

Let's merge this before people run into the same issue.


aminya:
Plots test pass locally! I wonder why it is not added to the pull request that the bot makes.

@timholy "Lost of methods" tests fail in the nightly. Has anything changed in nightly?

aminya:
I wonder if https://github.com/timholy/SnoopCompile.jl/pull/122/commits/f50b27a059af56bc910038779900d4b18e431833 actually fix the issue. 😞 This is a patch that we cannot send to our users unless we  find all the repositories that use SnoopCompile.yml

We should take #99 serious now. 🤦‍♂️

aminya:
@timholy Let's merge this before people run into the same issue. Having a pull request polluted with whitespace changes is better than not having the includer file. This is mostly a doc update now.

aminya:
@timholy Could you take a look at this?

timholy:
Hmm, messing with a repo's git attributes seems beyond the scope of SnoopCompile, and it seems like material better placed in PkgTemplates than here. Is there a more minimal fix possible?

aminya:
> Hmm, messing with a repo's git attributes seems beyond the scope of SnoopCompile, and it seems like material better placed in PkgTemplates than here. Is there a more minimal fix possible?

Yes, it should be in PkgTemplates too, but specifically for this case, we might need some sort of managing this. One way, I can think of is to have a `SnoopCompileBot.post_generation` function that runs in the Create_PR step, and we can do some stuff in it (fixing line-ending issues) or some compatibility things that may arise later.

timholy:
I'd like this to be as small a fix as possible for this bug, and then let's split the bot out into a separate package and you can do more heroic repo-surgery there.

timholy:
Or alternatively, if it really is hard to fix this without something big (notifying all existing users, whatever), then let's just do the split now and you can fix in 2.0?

aminya:
As I said already, I am OK with splitting the bot into a separate repository. We can do it at any time.

For later fixes, I want to add that `post_generation` function to refactor the Create_PR step in `SnoopCompile.yml`.

Currently, I have to error to inform the users about the faulty bash script. That only affects the users of multios/multiversion configuration.

timholy:
As mentioned above, I don't plan to merge this before we switch to 2.0. Changing the git settings on the repo by default is beyond the scope of this package. I recommend creating the new bot repo package soon.

PR #123, opened 2020-07-02
set tmin default to 0.001

aminya:
0.001 seems better than 0.0 at least for most of the situations.


timholy:
There are two views of this package:
- it's a diagnostic toool to help understand what's happening internally
- it's a tool to help reduce compilation time

For the latter, setting this to 0.001 instead of 0 seems like a better default. But for the former, this change is tantamount to lying by default. Since the name of the package is SnoopCompile and not ReduceCompileTimes, I think the former wins, especially since any snoop script can set `tmin` manually.

PR #127, opened 2020-08-03
Slim down to SnoopCompile and SnoopCompileCore

### LINKED ISSUE #77: Provide convenience utilities for invalidation
https://github.com/JuliaLang/julia/pull/35729 will make it possible to determine the causes of invalidation. It would be nice to organize these reports in a way that makes them more digestible to the user. It should be a fairly simple parsing problem, along with managing the state of the debug flag.

Helping users interpret the cause & cure might be more challenging; the kind of analysis in https://github.com/timholy/Revise.jl/issues/456#issuecomment-623024569 requires knowledge of many different aspect of Julia. Not sure what besides documentation one can do here.

@vtjnash, what do you think should be done with the `-- isequal(Method, Method)` lines? I'm inclined to drop them if they don't trigger a non-`--` invalidation, but perhaps I don't appreciate their full significance.

### LINKED ISSUE #99: SnoopCompileBot deprecations notice

### LINKED ISSUE #100: Update documentation to reflect the split
We should mention how to correctly use SnoopCompile for those who want to use the inner Core and Analysis packages directly, but in my opinion, we should state that the easiest and safest way to use SnoopCompile features is through SnoopCompileBot (offline or online) [because](https://github.com/timholy/SnoopCompile.jl/blob/master/SnoopCompileBot/src/snoop_bot.jl#L108) it loads the packages in the correct order, uses an external process that will be always clean, and adds some extra features like benchmarking, automatic precompile generation and handling.

I will need to address my last comment in #77 to make SnoopCompileBot a safe interface to **all** of the snooping operations including invalidations.

Copied from #98

### LINKED ISSUE #115: Versions in BotConfig and SnoopCompile.yml
In the docs it is highlighted that the versions specified in SnoopCompile.yml have to match the versions in `BotConfig`. In the .yml file I can use the versions
```
      matrix:
        version:   # NOTE: the versions below should match those in your botconfig
          - '1'
          - '1.3'
          - 'nightly'
```
where (I think) `'1'` corresponds to `'1.4.2'` and `'1.3'` to `'1.3.1'`.
However:
```julia
julia> BotConfig("Plots", version = ["1", "1.3", "nightly"]).version
3-element Array{VersionNumber,1}:
 v"1.0.0"
 v"1.3.0"
 v"1.6.0-DEV"
```
Does this mean I have to set the versions to `1.3.1` and `1.4.2` in both (or only BotConfig respectively) and update the versions manually with new julia releases, or is there another way to do this?

timholy:
This will be version 2.0 of SnoopCompile. It excises the bot, which should move to a separate repository. It also merges SnoopCompileAnalysis back into SnoopCompile itself.

Closes #118
Closes #121
Closes #122
Closes #77
Closes #99
Closes #100
Closes #115


aminya:
Thanks for this. I will move the repository to my account. 

I want to vote against merging Analysis and Core modules. The analysis module compiles parts of code that can make the snooping result inaccurate in some cases. For example, when you want to use SnoopCompile for snooping Pkg or some other core library. These deps can bring a lot of compiled code which decreases the accuracy of snooping for such core packages:
```
Cthulhu 
OrderedCollections 
```

timholy:
Core and "Analysis" are still separate. What I did was basically take Analysis's code and put it in SnoopCompile itself. So you can do all the snooping with `SnoopCompileCore`, and then just say `using SnoopCompile`. That gives you want you used to get with `SnoopCompileAnalysis`.

aminya:
> Core and "Analysis" are still separate. What I did was basically take Analysis's code and put it in SnoopCompile itself. So you can do all the snooping with `SnoopCompileCore`, and then just say `using SnoopCompile`. That gives you want you used to get with `SnoopCompileAnalysis`.

Oh, I see. That would work. 👍

aminya:
@timholy I want to move SnoopCompileBot to my account, but I have a question. Should I start a fresh git repository or there is a way to keep the git history for certain files?

PR #138, opened 2020-09-11
Use changes to Julia to have `@snoopi` report _all MethodInstances_ that are inferred during a single inference call

NHDaly:
Builds on https://github.com/JuliaLang/julia/pull/37535 to expose this information through SnoopCompileCore

### DIFF
--- a/SnoopCompileCore/src/snoopi.jl
+++ b/SnoopCompileCore/src/snoopi.jl
@@ -2,24 +2,36 @@ export @snoopi
 
 const __inf_timing__ = Tuple{Float64,MethodInstance}[]
 const __inf_callees__ = Dict{MethodInstance, Vector{MethodInstance}}()
+const __inf_callee_edges4__ = Dict{MethodInstance, Vector{Any}}()
 
 if isdefined(Core.Compiler, :Params)
     function typeinf_ext_timed(linfo::Core.MethodInstance, params::Core.Compiler.Params)
+        #Core.println("starting $linfo")
         empty!(Core.Compiler.__inference_callees__)
+        #empty!(Core.Compiler.__inference_callee_edges4__)
+        push!(Core.Compiler.__inference_callee_edges4__, (linfo, Tuple{MethodInstance,MethodInstance}[]))
         tstart = time()
         ret = Core.Compiler.typeinf_ext(linfo, params)
         tstop = time()
         push!(__inf_timing__, (tstop-tstart, linfo))
         __inf_callees__[linfo] = Core.Compiler.__inference_callees__
+        @assert Core.Compiler.__inference_callee_edges4__[end][1] == linfo
+        __inf_callee_edges4__[linfo] = pop!(Core.Compiler.__inference_callee_edges4__)[2]
+        #Core.println("push: ", Core.Compiler.__inference_callee_edges4__)
         return ret
     end
     function typeinf_ext_timed(linfo::Core.MethodInstance, world::UInt)
+        #Core.println("starting $linfo")
         empty!(Core.Compiler.__inference_callees__)
+        push!(Core.Compiler.__inference_callee_edges4__, (linfo, Tuple{MethodInstance,MethodInstance}[]))
         tstart = time()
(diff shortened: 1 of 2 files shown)


NHDaly:
I'm quite sure that I haven't gone about this in the best way. I'd love any suggestions or help if you are interested!!!! :) ❤️ !

Hopefully this will prove useful for others as well!

timholy:
Awesome that you're doing this. If I don't notice when you turn this into a real PR, just ping me.

NHDaly:
Thank you @timholy!!

-------

After some more thinking, together with @sacha0, i think we think that the approach we'd taken is for is probably _not_ the best way to go about measuring this.

Our next idea is to try a bit more like the other approaches, and just find a better place to inject the timings.
The reason i didn't think timings would work when i first looked at this, is that each function encapsulates the time for each subsequent function that it measures, so the timings aren't useful.

But then we realized if we were just able to report the timings in a nested structure, _like a profile_, they'd make sense! And then i remembered this is exactly what https://github.com/KristofferC/TimerOutputs.jl already does! 😁  So i think either duplicating that, or literally just using that, would get us what we want! As a first quick test, I just slapped TimerOutputs inside the compiler, and in this small test it seems indeed reasonable:
```julia
julia> module M
           i(x) = x+5
           i2(x) = x+2
           h(a::Array) = i2(a[1]::Integer) + i(a[1]::Integer) + 2
           g(y::Integer, x) = h(Any[y]) + Int(x)
       end;
WARNING: replacing module M.

julia> _to_ = TimerOutputs.TimerOutput();

julia> @code_typed M.g(2,3);   # Measure inference timee

julia> _to_ 
 ─────────────────────────────────────────────────────────────────────────────────────────────────
                                                          Time                   Allocations      
                                                  ──────────────────────   ───────────────────────
                 Tot / % measured:                     9.65s / 0.02%            887KiB / 30.7%    

 Section                                  ncalls     time   %tot     avg     alloc   %tot      avg
 ─────────────────────────────────────────────────────────────────────────────────────────────────
 MethodInstance for g(::Int64, ::Int64)        1   2.16ms   100%  2.16ms    273KiB  100%    273KiB
   MethodInstance for h(::Vector{Any})         1    766μs  35.5%   766μs    187KiB  68.7%   187KiB
     MethodInstance for i2(::Integer)          1    236μs  10.9%   236μs   42.6KiB  15.6%  42.6KiB
     MethodInstance for i(::Integer)           1    178μs  8.23%   178μs   42.6KiB  15.6%  42.6KiB
 ─────────────────────────────────────────────────────────────────────────────────────────────────
```
So i'll give an approach like that a go, next!! 😁

NHDaly:
(I've filed an issue in TimerOutputs.jl to see about using it to generate essentially profiles, like i want above 😊:
https://github.com/KristofferC/TimerOutputs.jl/issues/90)

NHDaly:
(fwiw, here's the git diff for what i'm trying right now):
```diff
diff --git a/base/compiler/typeinfer.jl b/base/compiler/typeinfer.jl
index a0abd97276..a8bec821b5 100644
--- a/base/compiler/typeinfer.jl
+++ b/base/compiler/typeinfer.jl
@@ -9,6 +9,11 @@ function typeinf(interp::AbstractInterpreter, result::InferenceResult, cached::B
 end

 function typeinf(interp::AbstractInterpreter, frame::InferenceState)
+    Main.TimerOutputs.@timeit Main._to_ Main.string(frame.linfo) begin
+        _typeinf(interp, frame)
+    end
+end
+function _typeinf(interp::AbstractInterpreter, frame::InferenceState)
     typeinf_nocycle(interp, frame) || return false # frame is now part of a higher cycle
     # with no active ip's, frame is done
     frames = frame.callers_in_cycle
```

timholy:
Cool. Not sure if you know this but `Revise.track(Core.Compiler)` may make this easier to test.

NHDaly:
Thanks @timholy -- I _didn't_ know that, and I appreciate the tip! 😁

PR #135, opened 2020-08-20
Add snoopl() to snoop time in LLVM optimization

NHDaly:
This uses the associated code added in https://github.com/JuliaLang/julia/pull/37136 to get its timings.

Adds a macro to SnoopCompileCore, `@snoopl`, which collects timings for the LLVM optimization of compilation, and adds functions to display and understand the resulting timings.

Co-Authored-By: @Sacha0 @vchuravy

PR #139, opened 2020-09-25
Expose changes to Julia via `@snoopi_deep` to report timing profile breakdown of _all MethodInstances_ that are inferred during type inference

NHDaly:
This builds on https://github.com/JuliaLang/julia/pull/37749 to expose the information through SnoopCompile.

This adds a macro to SnoopCompileCore `@snoopi_deep` which collects deep timings from Julia's type inference, and adds functions to display the resulting information as a profile.

TODO:
 - [ ] Add a docstring to `InferenceFrameInfo` in julia

-------

This is a second attempt at https://github.com/timholy/SnoopCompile.jl/pull/138.


NHDaly:
Early pass at SnoopI per-method-instance deep inference profile:
Screen Shot 2020-09-25 at 5 20 58 PM

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +94,57 @@ function _snoopi(cmd::Expr, tmin = 0.0)
...
+        end
+        finish_snoopi_deep()
+    end
+end
+
+macro snoopi_deep(cmd)
+    return _snoopi_deep(cmd)
+end
+
+function flatten_times(timing::Core.Compiler.Timings.Timing, tmin_secs = 0.0)
```
timholy, line 126: Does this need to be here or could it be moved to SnoopCompile? I regard SnoopCompileCore as "the minimum you need to collect the data" and anything not strictly required should live in SnoopCompile. (The more logic we have here, the more it becomes a source of potential artifacts wrt invalidation etc.)

NHDaly: Agreed. I think we should move it! :)

I had it here because in an earlier version i was calling this before returning; similar to `sort_timed_inf()`.

Thanks. Leaving this as a todo

aminya: Yes, definitely should be moved!

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +94,57 @@ function _snoopi(cmd::Expr, tmin = 0.0)
...
+function stop_deep_timing()
+    Core.Compiler.__set_measure_typeinf(false)
+    Core.Compiler.Timings.close_current_timer()
+end
+
+function finish_snoopi_deep()
+    return Core.Compiler.Timings._timings[1]
+end
+
+function _snoopi_deep(cmd::Expr)
```
aminya, line 110: Where is this function used?

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +94,57 @@ function _snoopi(cmd::Expr, tmin = 0.0)
...
+        try
+            $(esc(cmd))
+        finally
+            stop_deep_timing()
+        end
+        finish_snoopi_deep()
+    end
+end
+
+macro snoopi_deep(cmd)
```
aminya, line 122:
```suggestion
export @snoopi_deep
macro snoopi_deep(cmd)
```

SnoopCompileCore/src/snoopi.jl
NHDaly: Thanks Done.

SnoopCompileCore/src/snoopi.jl
NHDaly: I added the export to the top! :) Thanks

NHDaly:
This PR is still in draft stage, but i'm trying to clean it up now.
- [x] https://github.com/JuliaLang/julia/pull/37749 has been merged (thanks @timholy)!
- [x] I've added tests
- [x] I've added docstrings
- [ ] There is still some weirdness in trying to decide how best to print the names of the MethodInstances.
    - I think the (almost obviously) best choice is to print them via the built-in printing for MethodInstances, so e.g. `foo(::Integer)`, `foo(::Int)`, etc.
    - However, PProf currently strips the function arguments from the names of functions in the profile (🙄), so I've been using the TupleTypes directly instead.
        - But these are A) super ugly, and B) super long, and in e.g. ProfileView they go all the way across the screen so you can't read anything.
        - [ ] I plan to file an issue and/or PR on `pprof` itself to (at least have on option to) not modify the function names for Julia functions. Also, the current function regexes are unable to handle parentheses _around_ the function name (e.g. `(::MyStruct)(::Int)`), so we definitely need a PR to fix that.

aminya:
I would like to add deep snooping functionality to CompileBot.jl. I assume that all the configurations/inputs are the same as normal @snoopi, right?

NHDaly:
> I would like to add deep snooping functionality to CompileBot.jl. I assume that all the configurations/inputs are the same as normal @Snoopi, right?

I would say they're similar. The `commands` aspect should be the same, yeah:
https://github.com/timholy/SnoopCompile.jl/blob/1b6314962e3bb8faced5c15703e1a6fc2e1d2cf6/SnoopCompileCore/src/snoopi.jl#L123

And the `tmin` param can be passed to `SnoopCompile.flatten_times()` and `SnoopCompile.to_flamegraph()` instead of directly to `@snoopi_deep`.

timholy:
I'm supportive of your TODO list. I like the idea of fixing pprof and keeping the nicer printing. The last item (aggregation) sounds like it could also be added in a second PR, if you prefer (I'm fine with either).

NHDaly:
> I'm supportive of your TODO list. I like the idea of fixing pprof and keeping the nicer printing. 

Okay! Good news, the naming weirdness is settled in pprof-land:
I've asked about fixing this in `google/pprof`, and in the process I realized how we can work around it in `PProf.jl`: https://github.com/JuliaPerf/PProf.jl/pull/31

So I've updated this PR to use the MethodInstance name, and also filled out the StackFrame with more info. :)

> The last item (aggregation) sounds like it could also be added in a second PR, if you prefer (I'm fine with either).

Yeah, i like the idea of doing this in a follow-up! Good idea.

NHDaly:
Okay so I think this is looking good to go! :) I think this is ready to review.

I've updated to "Ready to Review". Thanks again for your help!

NHDaly:
(and i've renamed the PR now that it's ready to go!)

aminya:
You could add `fail-fast: false` to CI to allow the CI run on errors. 

https://github.com/timholy/SnoopCompile.jl/blob/93cefafa225de447283d05a0e66f062a019015d3/.github/workflows/ci.yml#L14

timholy:
Excited to see your progress, as you know I'm already benefiting from your work! I will try to review in the morning.

One thing I know I'm going to want, but not sure yet the best way to implement it, is to discover "who is calling the head of the next inference?" (`head = root.children[i]`). The reason I'm interested is that a break in the tree structure (starting a new head) indicates either toplevel or runtime dispatch. Sometimes there seem to be backedges but often not, so it's not entirely trivial short of inserting `@show stacktrace(backtrace())` into the callee. Currently, the best strategy I've devised is to collect all the leaves of the previous children and then run [`findcallers`](https://timholy.github.io/MethodAnalysis.jl/dev/#MethodAnalysis.findcallers) on them. Any better ideas? I'm happy to implement this part myself, I don't mean to suggest you need to do it for this PR.

NHDaly:
> I'm already benefiting from your work! 

🎉 

> "who is calling the head of the next inference?"

Yeah, interesting!

Starting a new top-level caller would happen (from what i understand) at runtime, after we've finished compiling and are now in the middle of executing, the code from (one of) the previous inference(s). So yeah, they're not connected at all.

And yeah, I expect there wouldn't be backedges set during dynamic dispatch, since you don't need to recompile the caller if the callee is updated (since it will _still_ be a dynamic dispatch).

But hrmm, yeah, interesting. I guess the best way to implement something like this would indeed be to walk the stacktrace at the first call to `typeinf` (if the parent is `ROOT`), and walk up until you hit something that's not type inference i guess? And then record that as the caller? Interesting idea.


> then run [`findcallers`](https://timholy.github.io/MethodAnalysis.jl/dev/#MethodAnalysis.findcallers) on them

Oooh this is cool! I hadn't seen this before. Neat! :)

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +86,74 @@ function _snoopi(cmd::Expr, tmin = 0.0)
...
+ 0.001618478 => Core.Compiler.Timings.InferenceFrameInfo(MethodInstance for rand!(::Random.MersenneTwister, ::Random.UnsafeView{Float64}, ::Random.SamplerTrivial{Random.CloseOpen01{Float64}, Float64}), 0x00000000000072e4, Any[], Any[Random.MersenneTwister, Random.UnsafeView{Float64}, Core.Const(Random.SamplerTrivial{Random.CloseOpen01{Float64}, Float64}(Random.CloseOpen01{Float64}())), Union{}, Union{}, Union{}, Union{}, Union{}, Union{}, Union{}])
+ 0.002289655 => Core.Compiler.Timings.InferenceFrameInfo(MethodInstance for _rand_max383!(::Random.MersenneTwister, ::Random.UnsafeView{Float64}, ::Random.CloseOpen01{Float64}), 0x00000000000072e4, Any[], Any[Random.MersenneTwister, Random.UnsafeView{Float64}, Core.Const(Random.CloseOpen01{Float64}()), Union{}, Union{}, Union{}, Union{}])
+ 0.093143594 => Core.Compiler.Timings.InferenceFrameInfo(MethodInstance for ROOT(), 0x0000000000000000, Any[], Any[])
```
NHDaly, line 140: Haha woah, these are really long. Sorry, i didn't realize because my editor doesn't wrap.

Is this example useful? Should I trim the output somehow?

KristofferC:
Could also be something like `@snoopi deep=true` (similar to `@code_warntype optimize=true`), but maybe they are different enough that it doesn't make sense to have them behind the same macro.

Project.toml
```diff
@@ -5,18 +5,23 @@ version = "2.1.1"
 
 [deps]
 Cthulhu = "f68482b8-f384-11e8-15f7-abe071a5a75f"
+FlameGraphs = "08572546-2f56-4bcf-ba4e-bab62c3a3f89"
+LeftChildRightSiblingTrees = "1d6d02ad-be62-4b6b-8a6d-2f90e265016e"
```
timholy, line 9: To minimize [compat] bumps, how about getting this from FlameGraphs instead of making it an independent dependency?
```julia
import FlameGraphs
using FlameGraphs: Node, addchild
```
below.

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +86,74 @@ function _snoopi(cmd::Expr, tmin = 0.0)
     end
 end
```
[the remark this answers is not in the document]
timholy: I think having it in a separate file, with a single `if` guard, makes a lot of sense.

> Also, how do you figure out which -DEV version of julia should be the minimum?

Demo:
```julia
tim@diva:~/src/julia-master$ contrib/commit-name.sh affbc2cec17939a113976345d3f5570f4bfde0d2
1.6.0-DEV.1301
```
When I remember I also like to link to the PR in a comment (it's hard to figure out which PR implemented a particular DEV number).

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +86,74 @@ function _snoopi(cmd::Expr, tmin = 0.0)
...
+        end
+        finish_snoopi_deep()
+    end
+end
+
+"""
+    timing_tree = @snoopi_deep commands
+
+Produce a profile of julia's type inference, containing the amount of time spent infering
+for every `MethodInstance` processed while executing `commands`.
```
timholy, line 118:
```suggestion
Produce a profile of julia's type inference, containing the amount of time spent inferring
every `MethodInstance` processed while executing `commands`.
```

SnoopCompileCore/src/snoopi.jl
timholy: Probably best. Maybe make it `InferenceFrameInfo(MethodInstance for fpsort!...)`?

We might also want to add a docstring for `InferenceFrameInfo`.

SnoopCompileCore/src/snoopi.jl
```diff
@@ -86,6 +86,74 @@ function _snoopi(cmd::Expr, tmin = 0.0)
...
+ 0.001618478 => Core.Compiler.Timings.InferenceFrameInfo(MethodInstance for rand!(::Random.MersenneTwister, ::Random.UnsafeView{Float64}, ::Random.SamplerTrivial{Random.CloseOpen01{Float64}, Float64}), 0x00000000000072e4, Any[], Any[Random.MersenneTwister, Random.UnsafeView{Float64}, Core.Const(Random.SamplerTrivial{Random.CloseOpen01{Float64}, Float64}(Random.CloseOpen01{Float64}())), Union{}, Union{}, Union{}, Union{}, Union{}, Union{}, Union{}])
+ 0.002289655 => Core.Compiler.Timings.InferenceFrameInfo(MethodInstance for _rand_max383!(::Random.MersenneTwister, ::Random.UnsafeView{Float64}, ::Random.CloseOpen01{Float64}), 0x00000000000072e4, Any[], Any[Random.MersenneTwister, Random.UnsafeView{Float64}, Core.Const(Random.CloseOpen01{Float64}()), Union{}, Union{}, Union{}, Union{}])
+ 0.093143594 => Core.Compiler.Timings.InferenceFrameInfo(MethodInstance for ROOT(), 0x0000000000000000, Any[], Any[])
+
+julia> fg = SnoopCompile.to_flamegraph(timing)
```
timholy, line 142: Why isn't this just another `flamegraph` method?

SnoopCompileCore/src/snoopi.jl
```diff
@@ -99,5 +167,11 @@ function __init__()
     end
     precompile(start_timing, ())
     precompile(stop_timing, ())
+
+    @assert precompile(Core.Compiler.Timings.reset_timings, ())
```
timholy, line 171: Are these actually required? Unlike `@snoopi` you're not calling `jl_set_typeinf_func`, right?

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
 ])
+
+# === @snoopi_deep helper functions ========================================================
+
+"""
+    flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
+
+Flatten the execution graph of Timings returned from `@snoopi_deep` into a Vector of pairs,
+with the exclusive time for each invcation of type inference within the compiler, sorted by
+the exclusive time, skipping any frames that took less than `tmin_secs` seconds.
```
timholy, line 391:
```suggestion
Flatten the execution graph of Timings returned from `@snoopi_deep` into a Vector of pairs,
with the exclusive time for each invocation of type inference, skipping any frames that took less than `tmin_secs` seconds. Results are sorted by time.
```

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+"""
+function flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
+    out = Any[]
```
timholy, line 394:
```suggestion
    out = Pair{Float64,Core.Compiler.Timings.InferenceFrameInfo}[]
```

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+end
+
+inclusive_time(t::InclusiveTiming) = t.inclusive_time
+
+function build_inclusive_times(t::Timing)
+    child_times = InclusiveTiming[
+        build_inclusive_times(child)
+        for child in t.children
+    ]
+    incl_time = t.time + sum(inclusive_time.(child_times); init=UInt64(0))
```
timholy, line 430:
```suggestion
    incl_time = t.time + sum(inclusive_time, child_times; init=UInt64(0))
```

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+Node(FlameGraphs.NodeData(ROOT() at typeinfer.jl:70, 0x00, 0:15355670))
+```
+
+NOTE: This function must touch every frame in the provided `Timing` to build inclusive
+timing information (`InclusiveTiming`). If you have a very large profile, and you plan to
+call this function multiple times (say with different values for `tmin_secs`), you can save
+some intermediate time by first calling [`build_inclusive_times(t)`](@ref), only once,
+and then passing in the `InclusiveTiming` object for all subsequent calls.
+"""
+function to_flamegraph(t::Timing; tmin_secs = 0.0)
```
timholy, line 467: As stated above, shouldn't this just be a new method of `FlameGraphs.flamegraph`?

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+    frontier = [timing]
+    while !isempty(frontier)
+        t = popfirst!(frontier)
+        exclusive_time = (t.time / 1e9)
+        if exclusive_time >= tmin_secs
+            push!(out, exclusive_time => t.mi_info)
+        end
+        for c in t.children
+            push!(frontier, c)
+        end
```
timholy, line 404:
```suggestion
        append!(frontier, t.children)
```

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+call this function multiple times (say with different values for `tmin_secs`), you can save
+some intermediate time by first calling [`build_inclusive_times(t)`](@ref), only once,
+and then passing in the `InclusiveTiming` object for all subsequent calls.
+"""
+function to_flamegraph(t::Timing; tmin_secs = 0.0)
+    it = build_inclusive_times(t)
+    to_flamegraph(it; tmin_secs=tmin_secs)
+end
+
+function to_flamegraph(to::InclusiveTiming; tmin_secs = 0.0)
```
timholy, line 472: This one can stay private since it's an internal method.

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+end
+# Special printing for Type Tuples so they're less ugly in the FlameGraph
+function frame_name(name, ::Type{TT}) where TT<:Tuple
+    io = IOBuffer()
+    Base.show_tuple_as_call(io, name, TT)
+    v = String(take!(io))
+    return v
+end
+
+# NOTE: The "root" node doesn't cover th whole profile, because it's only the _complement_
```
timholy, line 505:
```suggestion
# NOTE: The "root" node doesn't cover the whole profile, because it's only the _complement_
```

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+# of the inference times (so it's missing the _overhead_ from the measurement).
+# SO we need to manually create a root node that covers the whole thing.
+function max_end_time(t::InclusiveTiming)
+    return maximum(child.start_time + child.inclusive_time for child in t.children;
+                    init = UInt64(t.start_time + t.inclusive_time))
```
timholy, line 510: I'm not sure I understand the logic here...why is `start_time` relevant?

src/parcel_snoopi.jl
```diff
@@ -380,3 +380,150 @@ end
...
+import FlameGraphs
+
+using Base.StackTraces: StackFrame
+using LeftChildRightSiblingTrees: Node, addchild
+using Core.Compiler.Timings: Timing
+
+struct InclusiveTiming
+    mi_info::Core.Compiler.Timings.InferenceFrameInfo
+    inclusive_time::UInt64
+    start_time::UInt64
```
timholy, line 419: What is `start_time` used for? (More about this below.)

timholy (commented):
Looking very nice! I don't fully understand the `start_time` logic here but the rest seems sound. I've made a few localized suggestions but nothing that reshapes the structure of the PR.

src/parcel_snoopi.jl
NHDaly: Makes sense. I _do_ suggest in the documentation above that if you have a very large profile that you want to do interesting things with, that you can first call `build_inclusive_times(t)` and then pass in the _InclusiveTimings_ to this function and `flatten_times()`. Should I therefore rename this one to `flamegraph()` as well?

src/parcel_snoopi.jl
NHDaly: This is needed because the nodes in `FlameGraph` aren't just a duration: they are a _range_, from a start sample to an end sample:
```julia
julia> fg = SnoopCompile.to_flamegraph(timing)
Node(FlameGraphs.NodeData(ROOT() at typeinfer.jl:70, 0x00, 0:14424075))

julia> fg.child.child.child
Node(FlameGraphs.NodeData(i(x::Integer) at REPL[8]:2, 0x00, 168721:646740))
```
This (`0:14424075` vs `168721:646740`) is what allows the profile to have elements show up at the actual time they occurred, relative to eachother.

For example, in this image, `i(x::Int64)` only occurs at _runtime_, via dynamic dispatch, so you see a big gap in-between the first inference stack and it:

94317053-7fd11a00-ff53-11ea-9f7a-7e5ba7de0eef

Which comes from this:
```julia
julia> fg.child.sibling
Node(FlameGraphs.NodeData(i(x::Int64) at REPL[8]:2, 0x00, 11762162:11931975))
```

Unless I'm missing something, this `start_time` logic is required to get that to work, because in each node in julia's `Core.Compiler.Timings.Timing`, we're only keeping the total exclusive time and the start time:
https://github.com/NHDaly/julia/blob/f6f4705f943abebcb4eaa50bb78e9cb9f46d0c43/base/compiler/typeinfer.jl#L48-L56

So to recompute the spans needed for the `range` in the FlameGraph, we need to keep the start time, and then sum the inclusive time with the inclusive times of all children, which gives the range, as `t.start_time : (t.start_time + t.inclusive_time)`.

Does that answer your question? :)

src/parcel_snoopi.jl
NHDaly: I've explained why start_time is relevant _in general_ in the review comment above.

In this particular function, it's used to calculate an artificial "<root>" node that extends long enough to cover all of the top-level inference calls. Because without this, I was hitting a problem where the exclusive time of the root node, plus all of the inclusive times of its children, didn't _quite_ cover every nanosecond of the last top-level inference call. I think this is because there's a tiny bit of overhead in the process of starting and stopping the inference which we explicitly exclude from our measurements (in https://github.com/JuliaLang/julia/pull/37749), and that adds up over all the stacks, so that the ROOT() node doesn't quite extend long enough to cover the end time of its last child, and this violates an assertion within `FlameGraphs`.

So this function is used to compute the maximum **end time** by adding each child's start time to its duration. Hopefully that makes sense now?

------

However, after explaining the above, I've now realized that the `children` array is actually sorted by start_time already, since we do depth first search and push_back each node as soon as it's finished. So I guess I could simplify this to simply be
```julia
    last_node = length(t.children) > 0 ? t.children[end] : t
    return last_node.start_time + last_node.inclusive_time
```
Do you think this is clearer?

SnoopCompileCore/src/snoopi.jl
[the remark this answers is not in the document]
NHDaly: Done. Thanks

SnoopCompileCore/src/snoopi.jl
NHDaly: Also, another reason to keep these separate is that they have entirely non-overlapping keyword arguments, so it doesn't really feel like another method in the same function...

What do you think?

NHDaly:
> Could also be something like `@snoopi deep=true` (similar to `@code_warntype optimize=true`), but maybe they are different enough that it doesn't make sense to have them behind the same macro.

@KristofferC - sorry i never responded to this. I had the same idea at first, but since this returns an entirely different data structure, and is used with entirely different helper functions, it doesn't seem to make sense to me to put them in the same macro.

NHDaly:
Alright, tests are passing, and I'm finished applying changes! :)

Please let me know what you think about the `to_flamegraph` / `flamegraph` message, and then otherwise I think this is good to go! 🎸

SnoopCompileCore/src/snoopi.jl
timholy: `@snoopi` rips out the function that's currently being used to infer code and replaces it with something new. Unfortunately, if in the "in between" state you need to run inference, you're hosed. So precompilation is used to make sure that never happens. Try deleting those from `@snoopi` and see what happens (I don't remember the answer anymore, but I am almost certain there is some horrible,session-catastrophic error), then try the same thing in your PR.

src/parcel_snoopi.jl
timholy: Very nicely, thanks!

src/parcel_snoopi.jl
timholy: That does seem a bit clearer.

timholy:
This looks basically ready to go. The one thing I'd still recommend (but not insist on, if you disagree) is making the "public" methods `flamegraph` rather than `to_flamegraph`. Here are the reasons:
- there are two ways to think of a flamegraph: (1) a tool for visualizing runtime costs (this does *not* fit that) (2) a data structure in which parent-child relationships are represented in two dimensions, where the horizontal axis is used to represent a timing dimension (this *does* fit that). 
- I think of `flamegraph` almost as a constructor (there is no `FlameGraph` type in `FlameGraphs`, because it's just the root node of a tree), and we have constructors that take very diverse inputs and return the same type of object. For example, `Array{Int}(undef, 3)` and `Array{Int}(1:3)` are very different beasts indeed, yet we call them the same thing because they produce the same output type.

If you do agree that it can safely be another method of `flamegraph`, perhaps we should even consider having SnoopCompile re-export than name.

timholy:
Also, PRs like this are a testament to both your own excellent work and Julia's amazing composability: if you deleted all the documentation (no, please don't!) this is a remarkably small PR for a huge gain in functionality. Pretty incredible.

SnoopCompileCore/src/snoopi.jl
NHDaly: Oh, I see! :) Thanks. I assumed they were there so that these functions don't show up in trace that's being recorded. That makes sense! I'll see what happens if I delete them. 👍

timholy: Oh, good point. If they are annoying then they are fine to precompile. Since they were initially in the same place as the others I thought you had a different purpose.

src/parcel_snoopi.jl
timholy: Sure, that would be fine.

src/parcel_snoopi.jl
NHDaly: Done in latest push. Thanks!

NHDaly:
Thanks @timholy for the kind words! ❤️ 

Agreed regarding how amazing it is that we could add such interesting functionality in such small PRs. 😊 

---

Per your suggestion, I've changed to implement `flamegraph()`, and re-exported it. Thanks!!

I believe this is ready for final review!

SnoopCompileCore/src/SnoopCompileCore.jl
```diff
@@ -11,6 +11,9 @@ if VERSION >= v"1.2.0-DEV.573"
     include("snoopi.jl")
 end
 
+if VERSION >= v"1.6.0-DEV.1190"  # https://github.com/JuliaLang/julia/pull/37749
```
timholy, line 14: I swear I commented about this somewhere but I can't find it; it's fine with me if you want to load the parceling even on earlier Julia versions (your choice).

src/parcel_snoopi_deep.jl
```diff
@@ -0,0 +1,147 @@
...
+julia> ProfileView.view(fg);  # Display the FlameGraph in a package that supports it
+
+julia> fg = flamegraph(timing; tmin_secs=0.0001)  # Skip very tiny frames
+Node(FlameGraphs.NodeData(ROOT() at typeinfer.jl:70, 0x00, 0:15355670))
+```
+
+NOTE: This function must touch every frame in the provided `Timing` to build inclusive
+timing information (`InclusiveTiming`). If you have a very large profile, and you plan to
+call this function multiple times (say with different values for `tmin_secs`), you can save
+some intermediate time by first calling [`build_inclusive_times(t)`](@ref), only once,
```
timholy, line 78:
```suggestion
some intermediate time by first calling [`SnoopCompile.build_inclusive_times(t)`](@ref), only once,
```

timholy (approved):
Merge at will!

SnoopCompileCore/src/SnoopCompileCore.jl
NHDaly: Ah, that was on the other PR. I don't think that will work on this PR, because the parcelling depends on the changes in julia. 😁

NHDaly:
😁 Thanks! :) I don't have merge permissions, so...... Merge at will! 😁

timholy:
> I don't have merge permissions

You should have gotten an email about this earlier today...let me know if you can't find it and I'll try to resend.

NHDaly:
Hehe oh, you're right! 😁 Thanks! :) Amazing. Really appreciate it! ✔️ 🥳 

Do you have any preference regarding Merge vs Squash? I'm inclined to Squash because there were a lot of commits here.

NHDaly:
Oh, also, I guess I should do a version bump here.

Is this a minor or patch change? (Or major?) I'm inclined to say minor? And do you think it's okay to release both `@snoopl` (#135) and `@snoopi_deep` (this PR) in the same new version?

timholy:
Yes to squash, yes to minor bump. Before registering, we might want to work on the Documenter docs a bit to advertise this functionality.

timholy:
I was going to let you do the honors, @NHDaly, but I'm eager to use this. Thanks again!

NHDaly:
Hehe, eep! I'm sorry -- I was excited to do the honors, too, but I felt like it probably needs more documentation, and then I got carried away doing some analyses now that we have these tools instead of writing the docs! 🙈 

Thanks for the merge, @timholy! :) I'll try to follow up with docs soon! <3

timholy:
I may be able to help too. Lots going on so it will be a bit back-burner but it's already helping, so let's at least work with it in master.

PR #140, opened 2020-10-01
Deprecate SnoopCompileBot in favor of CompileBot

aminya:
This deprecates SnoopCompileBot in favor of CompileBot. 

Because SnoopCopmileBot was in a sub-directory of this repository, changing the URL is not possible due to git limitations. I decided to rename the package.

I would appreciate it if you can merge this and register a new version for SnoopCompileBot. Then, remove the directory when the registration is done.

https://github.com/JuliaRegistries/General/pull/22253
https://github.com/JuliaRegistries/General/pull/19557
https://github.com/JuliaRegistries/General/pull/21615

### DIFF
--- a/SnoopCompileBot/Project.toml
+++ b/SnoopCompileBot/Project.toml
@@ -1,0 +1,13 @@
+name = "SnoopCompileBot"
+uuid = "1d5e0e55-7d74-4714-b8d8-efa80e938cf7"
+author = ["Amin Yahyaabadi <aminyahyaabadi74@gmail.com>"]
+version = "2.0.0"
+
+[deps]
+CompileBot = "877fb49a-56b0-480a-9049-319c0c353852"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
+
+[compat]
+CompileBot = "< 2"
+Reexport = "0.2"
+julia = "1"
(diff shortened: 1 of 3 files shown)


timholy:
Happy to do this, but just to make sure: where precisely is the problem? Are you talking about the registry? (I don't think that should be a problem, but you will have to submit a manual PR to General.) Or github transfer & ownership? Something else?

aminya:
Yes. The issue is in the registration. The general registry stores one URL for the whole package. That means the git-shas specified there rely on that URL. When I made the new package, I started fresh (I did not clone this repository). That means the old git-shas will not be valid anymore.

The long-term solution that I see is to extend Pkg so it stores URL for each version, instead of having one single URL for the whole thing. 

Another thing is that I had an idea to make an automated bot for PackageCompiler. I decided to see this issue as an opportunity for extending SnoopCopmileBot to CompileBot.

aminya:
I forgot to say that we should wait until the new package gets registered: https://github.com/JuliaRegistries/General/pull/22253

timholy:
If people only want the latest-and-greatest, I don't see a problem with just changing the URL. But perhaps you're wanting to maintain the availability of older versions too, in which case this makes sense.

As long as you're happy about the name change, then of course it's fine, just wanted to make sure you weren't giving up for the wrong reasons.

timholy:
Just ping me when you want me to merge this.

aminya:
@timholy CompileBot is now registered. This should be ready if no error happens.

timholy:
Not sure if https://github.com/JuliaRegistries/General/pull/23132 should have also included SnoopCompileBot? If necessary, please register a new version.

PR #151, opened 2020-11-25
snoopi_deep: check that mi.def is a Method

timholy:
The `def` field of `Core.MethodInstance` is typed as
`Union{Method,Module}`. Consequently, we should not assume that it's
a `Method`.

Unfortunately I've not been able to come up with a simple test
that fails without this change, though I have seen failures
in real-world usage.


timholy:
Let's see if this passes with #153 merged.

PR #152, opened 2020-11-25
Add accumulate_by_method

timholy:
This analysis tool is useful to detect methods that get a crazy number
of specializations. These are candidates for refactoring into smaller,
more independent chunks or nospecialize annotations.

### DIFF
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -13,7 +13,9 @@ Flatten the execution graph of Timings returned from `@snoopi_deep` into a Vecto
 with the exclusive time for each invocation of type inference, skipping any frames that
 took less than `tmin_secs` seconds. Results are sorted by time.
 
-`ROOT` is a dummy element whose time corresponds to the total time for the operation, not just inference time.
+`ROOT` is a dummy element whose time corresponds to the sum of time spent outside inference. It's
+the total time of the operation minus the total time for inference. You can run `sum(first.(result[1:end-1]))`
+to get the total inference time, and `sum(first.(result))` to get the total time overall.
 """
 function flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
     out = Pair{Float64,Core.Compiler.Timings.InferenceFrameInfo}[]
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -12,6 +12,8 @@ const flamegraph = FlameGraphs.flamegraph  # For re-export
 Flatten the execution graph of Timings returned from `@snoopi_deep` into a Vector of pairs,
 with the exclusive time for each invocation of type inference, skipping any frames that
 took less than `tmin_secs` seconds. Results are sorted by time.
+
+`ROOT` is a dummy element whose time corresponds to the total time for the operation, not just inference time.
 """
 function flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
     out = Pair{Float64,Core.Compiler.Timings.InferenceFrameInfo}[]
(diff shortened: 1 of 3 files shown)


timholy:
Test failure fixed by #153

src/parcel_snoopi_deep.jl
```diff
@@ -24,7 +26,45 @@ function flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
...
+function accumulate_by_method(pairs::Vector{Pair{Float64,Core.Compiler.Timings.InferenceFrameInfo}}; tmin_secs = 0.0)
+    tmp = Dict{Method,Float64}()
+    for (t, info) in pairs
+        m = info.mi.def
+        if isa(m, Method)
+            tmp[m] = get(tmp, m, 0.0) + t
+        end
+    end
+    return sort([t=>m for (m, t) in tmp if t >= tmin_secs]; by=first)
 end
```
NHDaly, line 68: This is perfect! 👌!

I should've realized we needed this function a while ago -- I've been just reusing this line from my julia REPL history 😬 !:
```julia
julia> ms = let ms = DefaultDict{Any,Float64}(0)
           for (t,mi_info) in times
               ms[mi_info.mi.def] += t
           end
           ms
       end;

julia> methods_sorted = sort([v=>k for (k,v) in ms], by=first)
```

It looks exactly identical to what you've done here 👍 👌

src/parcel_snoopi_deep.jl
```diff
@@ -24,7 +26,45 @@ function flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
...
+6-element Vector{Pair{Float64, Method}}:
+ 0.000590579 => _copyto_impl!(dest::Array, doffs::Integer, src::Array, soffs::Integer, n::Integer) in Base at array.jl:307
+ 0.000616788 => partition!(v::AbstractVector{T} where T, lo::Integer, hi::Integer, o::Base.Order.Ordering) in Base.Sort at sort.jl:578
+ 0.000634394 => sort!(v::AbstractVector{T} where T, lo::Integer, hi::Integer, ::Base.Sort.InsertionSortAlg, o::Base.Order.Ordering) in Base.Sort at sort.jl:527
+ 0.000720815 => Vector{T}(::UndefInitializer, m::Int64) where T in Core at boot.jl:448
+ 0.001157551 => getindex(::Type{T}, x, y, z) where T in Base at array.jl:394
+ 0.046509861 => ROOT() in Core.Compiler.Timings at compiler/typeinfer.jl:75
+```
+
+`ROOT` is a dummy element whose time corresponds to the total time for the operation, not just inference time.
```
NHDaly, line 57: See above comment :)

src/parcel_snoopi_deep.jl
```diff
@@ -12,6 +12,8 @@ const flamegraph = FlameGraphs.flamegraph  # For re-export
 Flatten the execution graph of Timings returned from `@snoopi_deep` into a Vector of pairs,
 with the exclusive time for each invocation of type inference, skipping any frames that
 took less than `tmin_secs` seconds. Results are sorted by time.
+
+`ROOT` is a dummy element whose time corresponds to the total time for the operation, not just inference time.
```
NHDaly, line 16: Ah, actually, `ROOT` is the _complement to_ inference time. It's the total time _minus_ inference time. It's like the "exclusive time" for the top-level node.

Sorry if that's confusing!

NHDaly: It looks like the only place I've documented that is here, and it's pretty poor:
https://github.com/JuliaLang/julia/blob/aee4c26ae4e5c3209a0a4962de813a34f78d5c9c/base/compiler/typeinfer.jl#L73-L83

Thanks for adding this line to these functions! 👍 Good idea

NHDaly: Maybe something like this would clarify things?:

```suggestion
`ROOT` is a dummy element whose time corresponds to the sum of time spent outside inference. It's
the total time of the operation minus the total time for inference. You can run `sum(first.(result[1:end-1]))`
to get the total inference time, and `sum(first.(result))` to get the total time overall.
```

NHDaly: Also also, i've been wanting to redo the flamegraph a bit to actually put the `ROOT()` nodes back into the flamegraph, inbetween the inference nodes, which I think will help clarify this? :)

NHDaly (commented):
Awesome Tim! :) Thanks, this is nice.

src/parcel_snoopi_deep.jl
timholy: > ROOT is the complement to inference time

Thanks. I got this correct in the blog post but I must not have noticed/remembered this yet when creating this PR. I like your suggested change.

timholy: > actually put the ROOT() nodes back into the flamegraph

That does seem like it would help, though with the combination of my 3 PRs I now rarely look at the flamegraph. That might be for reasons of performance though, and you've fixed that, so this seems worth having.

src/parcel_snoopi_deep.jl
timholy: One slight difference is that you allow `m` to be a `Module`. If we include modules then we might need to reconsider the name.

timholy (commented):
Sounds like we're mostly in accord here, just a couple of details to hash out.

src/parcel_snoopi_deep.jl
NHDaly:
```suggestion
`ROOT` is a dummy element whose time corresponds to the sum of time spent outside inference. 
```
I copied this from the above comment. It seems reasonable to me to limit this one to just one line, since it's explained more in `flatten_times()`, which you have to call before you can call this one. :)

src/parcel_snoopi_deep.jl
NHDaly: oh good point. I hadn't realized I think that `m` could be a Module...

I'm not sure what's best. I guess if one of the MethodInstances where `m` is a `Module` _was_ accounting for most of the time, I'd be sad if we missed it because we were filtering those out! So maybe we should include them too?

I dunno if we really need to change the name, since it's _usually_ a method? Or maybe we could call it `accumulate_by_definition()`, to be more inclusive?

src/parcel_snoopi_deep.jl
```diff
@@ -24,7 +28,45 @@ function flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
         end
         append!(frontier, t.children)
     end
-    return sort(out; by=tl->tl[1])
+    return sort(out; by=first)
+end
+
+"""
+    accumulate_by_method(pairs; tmin_secs = 0.0)
```
NHDaly, line 35: Also, can we add this as a reference to the docstring on `@snoopi_deep`?:
https://github.com/timholy/SnoopCompile.jl/blob/2f88978f6b7f691b3d46c515f3281d24c7342c15/SnoopCompileCore/src/snoopi_deep.jl#L35-L38

src/parcel_snoopi_deep.jl
NHDaly: > I now rarely look at the flamegraph. That might be for reasons of performance though, and you've fixed that, so this seems worth having.

Also, if you haven't already tried it, I'd _very much_ recommend trying to display the flame graph in PProf.jl, via `pprof(fg)`. In my opinion it's a _really_ nice experience to see all of the method instances laid out with names visible, and it allows searching and filtering, all of which really help with digging into performance.

Also, the graph view is useful if we support aggregated views, as is done in #160. :)

NHDaly (approved):
👍 I'll leave you to address the remaining comments as you see fit :) Thanks @timholy!

src/parcel_snoopi_deep.jl
NHDaly: :) I've opened a PR to add PProf as one of the recommended viewers for FlameGraphs, since i realized it wasn't there yet! :)
https://github.com/timholy/FlameGraphs.jl/pull/32

src/parcel_snoopi_deep.jl
timholy: Yes. I'm a little unhappy that the docstrings (really, Example sections) I've added are so long. I think I'll refactor that to include an example in `@snoopi_deep` that covers all the use cases and then refer to it (without repeating it) from each individual function. But it will be easiest to merge them all and then fix the docstrings, just so they can be harmonized. So don't worry too much about the Example section for now.

src/parcel_snoopi_deep.jl
timholy: OK, now the logic is:
- if `mi.def` is a Method, aggegation makes sense and use the Method as the key of the dict
- if `mi.def` is a module, it's perfectly possible for there to be multiple thunks from the same module. Hence, we just preserve `mi`.

timholy:
Now this is called `accumulate_by_source`, and I'll add a second method in #159. That will reduce the number of exports by 1, whew.

Issue #150, opened 2020-11-24
Too many coffee breaks while building the flamegraph


timholy:
I am *loving* `@snoopi_deep` but building the flamegraph can take minutes. All the time appears to be in this call: https://github.com/timholy/SnoopCompile.jl/blob/bf52c6bc3494905039f90ad6ea8f9fdf1c24202f/src/parcel_snoopi_deep.jl#L114

Can this be sped up? Or do we have to consider developing a lazy-string that will be shown only when the user hovers over it in ProfileView?

NHDaly:
Yeah, I've noticed this too. Apparently this is super duper slow.

> Or do we have to consider developing a lazy-string that will be shown only when the user hovers over it in ProfileView?

I'd rather speed it up, because with `PProf` the whole thing is exported to disk at once, and then all the names are displayed in the profile.

In this branch (i haven't opened a PR yet, but i think it's probably ready to), I've added other display modes, to aggregate by Method or Function, and also to allow displaying if the method instance is being specialized with `Const`s:
https://github.com/timholy/SnoopCompile.jl/compare/nhd-snoopi_deep-flamegraph-printing

Basically all of the other display modes print much faster, including `slottypes` (which displays the consts), so I think it certainly could be sped up. I think it's very likely that just no one has taken the time to optimize this, because until now no one was printing tens of thousands of MethodInstances in a loop 😅 

So I think either we could make our own printing mechanism, or we could try to improve that one? I'm in favor of trying to improve the printing in Base, if possible?

NHDaly:
(BTW, in case it's helpful: for now, I've been always just setting `tmin_secs = 0.01` to prevent the flamegraph building from taking too long.)

NHDaly:
😅 haha oh wow: I ran `@snoopi_deep @eval flamegraph(timings)` to find out where all that time is going, and 😅 half the time is in _inferring_ `frame_name` - I think probably I should just change the type tuple from a compile-time argument to a runtime argument! Lemme give that a shot
<img width="1309" alt="Screen Shot 2020-11-27 at 12 04 20 PM" src="https://user-images.githubusercontent.com/1582097/100471673-bf79c680-30a8-11eb-8bf9-faa5982e8bd4.png">

NHDaly:
(Here's the unaggregated version - the one you're used to seeing):
[...]

PR #158, opened 2020-11-27
Dramatically improve `flamegraph(timing)` performance

NHDaly:
Move method instance's specTypes Type Tuple from a compile-time argument
to a runtime argument, to prevent compiling a specialization for every
_value_ in the data!

I hadn't meant to write it that way in the first place, I had just done
it absent-mindedly, because `specTypes` is a Type, so I hadn't thought
about moving it back into the value domain.

We used `@snoopi_deep` to find out the problem and fix this performance
problem! It's neat to see it profiling itself! :tada:

Fixes https://github.com/timholy/SnoopCompile.jl/issues/150.

### DIFF
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -110,10 +110,16 @@ function frame_name(mi::Core.Compiler.MethodInstance)
 end
 # Special printing for Type Tuples so they're less ugly in the FlameGraph
 function frame_name(name, @nospecialize(tt::Type{<:Tuple}))
-    io = IOBuffer()
-    Base.show_tuple_as_call(io, name, tt)
-    v = String(take!(io))
-    return v
+    try
+        io = IOBuffer()
+        Base.show_tuple_as_call(io, name, tt)
+        v = String(take!(io))
+        return v
+    catch e
+        e isa InterruptException && rethrow()
+        @warn "Error displaying frame: $e"
+        return name
+    end
 end
 
 # NOTE: The "root" node doesn't cover the whole profile, because it's only the _complement_
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -109,9 +109,9 @@ function frame_name(mi::Core.Compiler.MethodInstance)
     frame_name(mi.def.name, mi.specTypes)
 end
 # Special printing for Type Tuples so they're less ugly in the FlameGraph
-function frame_name(name, ::Type{TT}) where TT<:Tuple
+function frame_name(name, @nospecialize(tt::Type{<:Tuple}))
     io = IOBuffer()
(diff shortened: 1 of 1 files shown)


NHDaly:
Also, in case you're curious:
Moving from `::Type{TT}) where TT<:Tuple` to `tt::Type{<:Tuple})` alone reduced the time by a lot, and it prevented the redundant specializations of `show_tuple_as_call()`. But there were still a bunch of specializations of the actual function itself, each one still compiled a new call to `current_logger_for_env` (based on other error handling i've added in a subsequent commit, which was there when i was profiling -- sorry for the confusion), so I also added a `@nospecialize` to prevent that from happening, which dropped the rest of the time way down. :)

NHDaly:
This will make this a lot more usable. Thanks for bringing it up, @timholy ! :)

timholy:
Ah, I shouldn't have squashed, the two commits were meaningful. Next time...

Thanks for this elegant fix! I hadn't thought to profile *this* package.

NHDaly:
> Ah, I shouldn't have squashed, the two commits were meaningful. Next time...

:) eh, they're in the PR history in github anyway. 

> Thanks for this elegant fix! I hadn't thought to profile this package.

Hehe, yeah, right? me neither, until now!
👍 thanks! :)

timholy:
BTW, to address the need for docs on `@snoopi_deep` I am writing up a blog post and making you a co-author :smile:. It uses DataFrames as a test case. DataFrames is maintained by sophisticated developers who have been vigilant about compile latency; it's a nice example where most  things that could have been mined by `@snoopi` have already been mined, yet there are new things for `@snoopi_deep` to discover. I'll submit this blog post soon as a PR and CC you, but it could use another 24hrs of polish before I submit it.

As usual, documenting something helps discover all sorts of additional needs. I'm in the process of fleshing out the `@snoopi_deep` utility API. I've just flagged a few PRs that I'd like your feedback on, to see whether you agree that these utilities are useful for analysis and whether there are improvements to be made.

NHDaly:
😍 😍 😍  That sounds -amazing-!! Thanks @timholy! I was hoping/planning to eventually get around to doing that too, but we have _so much_ technical debt when it comes to compilation latency, I think I could spend the next 3 years just trying to improve compilation latency of the code already written at RelationalAI, if I was doing it alone! 😬  So we've been pretty busy trying to find the major low-hanging fruit, but once that's done, and we've validated the workflow, I wanted to turn to educating the rest of our coworkers to use these tools themselves. So having a blog post will be amazingly helpful! ❤️!!

:) Looking forward to reading it!

timholy:
#157 should probably be the next one on the queue; I haven't updated #159 because there's no point in fixing merge conflicts for two PRs that will also conflict whenever the first one is merged.

PR #159, opened 2020-11-27
Add `inference_triggers`

timholy:
This adds a lot of exports, but the analysis of the callees and callers at breaks
in inference chains is fairly complicated. These utilities streamline the analysis.

This requires https://github.com/JuliaLang/julia/pull/38566


timholy:
Note when I fix merge conflicts I'll rename `accumulate_by_callsite` to `accumulate_by_source`, making it another method of the function introduced in #152.

PR #160, opened 2020-11-27
Add options to flamegraph() to display as methods, functions, or as slottypes (printing Consts)

NHDaly:
Currently, each MethodInstance is displayed individually, but this adds other options to display them. If you display them as Methods, or Functions, this has the effect of aggregating the results. If you display them as `slottypes`, it shows if a MethodInstance was specialized with constants.

This also covers aggregation if you use SnoopCompile together with a flamegraphs renderer that supports displaying aggregations, like PProf.jl, and it shows a simplified view if you use it with a linear-ordered Flamegraph renderer like ProfileView.jl.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
           - '1.0'
           - '1.2'
           - '1.4'
+          - '1'
           - 'nightly'
         os:
           - ubuntu-latest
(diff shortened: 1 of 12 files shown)


timholy:
See https://github.com/JuliaLang/julia/pull/38596 for related work.

NHDaly:
Ooh, thanks ~~ that looks really great! :) Yes, we should definitely base this PR on top of that work! 👍 👍 

We probably won't need the slottypes option to print the Consts, then, since we can just use what you've added 👍  :)

NHDaly:
Also, I like what you did here with the generators. I hadn't noticed that, we should probably copy that into this as well when generating the flamegraphs!
https://github.com/JuliaLang/julia/blob/3b4b512eae901587347176c19674907a953cd0bd/base/show.jl#L1006-L1008

timholy:
xref #175

PR #179, opened 2020-12-20
Allow `flatten_times` to return unsorted

timholy:
When analyzing the overhead of specialization, sometimes you want
to distinguish the first run (or first run with concrete signature)
from later runs. Consequently it can be useful to extract the
flattened times in the order in which inference happened.

This also switches to a depth-first algorithm for searching the graph,
as required to support `sorted=false`.


NHDaly (commented):
This is a great idea, yeah! 👍 👍 

Thanks! :)

PR #180, opened 2020-12-21
Add Inclusive Time to results from `flatten_times()`

NHDaly:
This updates `flatten_times()` to return _both_ the exclusive _and_ the inclusive time for every frame in the trace.

To do this, we first compute the inclusive times from the profile and then use that for flatten_times. This _does_ make the function take longer, but it makes the results so much more valuable that I think it's worth it.

This gives you results that feel similar to the results from `@snoopi`, but for every level in the profile, not just the top-level.

I applied the same change to `accumulate_by_source()` while I was there.

### DIFF
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -11,85 +11,6 @@ isROOT(mi_info::InferenceFrameInfo) = isROOT(mi_info.mi)
 isROOT(mi::MethodInstance) = mi === Core.Compiler.Timings.ROOTmi
 isROOT(m::Method) = m === Core.Compiler.Timings.ROOTmi.def
 
-"""
-    flatten_times(timing::Core.Compiler.Timings.Timing; tmin_secs = 0.0)
-
-Flatten the execution graph of Timings returned from `@snoopi_deep` into a Vector of
-tuples, containing the exclusive time and inclusive time for each invocation of type
-inference, skipping any frames that took less than `tmin_secs` seconds. Results are
-sorted by exclusive time.
-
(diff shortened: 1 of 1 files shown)


NHDaly:
I wrote this change on Friday evening, but hadn't opened a PR for it yet, and it now conflicts with your most recent changes, like https://github.com/timholy/SnoopCompile.jl/pull/179 😅

But i think it should merge in well from what I can tell! :) 👍

timholy:
Welcome back!

Lots of changes that may need to be addressed. For instance, now there isn't a `build_inclusive_times` anymore; it's just a call to the `InclusiveTiming` constructor.

timholy:
I agree that there's a lot of redundancy in the data with the current scheme. But there are also disadvantages in combining them.  As it is, now you can do an analysis on either inclusive or exclusive times by variable selection (https://github.com/JuliaPlots/Makie.jl/issues/792#issuecomment-748625513), with this change you'd have to have some other way of specifying which field you want to apply each analysis to. For example, if you sort them, you have to pick one or the other, and re-sort if you want to compare them. If you have them in two separate trees, it might make an easier API?

But the redundancy is a problem, so I am open to making this kind of change.

timholy:
On the other hand...if you've done some analysis, like pulling out frames from a particular module or method, and then want to switch between inclusive vs exclusive times, that would favor your proposal.

It seems worth trying. What about making it a mutable structure with a bool that sets which "view" we're using currently? We'll need this both for the tree structure and individual frames, so perhaps it makes sense to come up with an analog of `InferenceFrameInfo` that includes the timing and the bool and then redefine the tree structure to contain it.

timholy:
I've gone ahead and finished this off in #191.

PR #183, opened 2020-12-22
Skip higher-order funcs for inference triggers

timholy:
When identifying the caller of an inference-triggering callee,
sometimes it's nice to skip over higher order functions.
For instance in a `do`-block, you might prefer to attribute
any inference trigger in the anonymous function to the parent
scope that implemented the `do`-block rather than the
higher-order function used by the `do`-statement.


timholy:
Test failure is due to a timing shift and unrelated to this PR. Will fix in a separate PR.

PR #186, opened 2020-12-23
`ascend` integration and docs cleanup

timholy:
This provides integration with Cthulhu's `ascend`, making it
significantly easier to analyze inference triggers.

The docstrings had also gotten out of date, and this updates them
while also allowing more compact docstrings by moving definitions to a
set of demos.

Finally, this adds an "orientation" docstring to SnoopCompile itself.
These can be helpful in orienting users to the API.


timholy:
Because this will be a big generator of conflicts (due to moving code around), I'm merging ASAP so that resolution of #180 can be done against the latest.

PR #189, opened 2020-12-26
Improve accuracy & speed of known_type

timholy:
This gets `known_type` out of the business of `eval`ing type names and
uses the internal machinery instead.


timholy:
Seems to be intermittent.

PR #190, opened 2020-12-27
Rename snoopi_deep -> snoopi_tree

timholy:
I don't think this is a big win, but "deep" seems pretty nonspecific.
I personally find "tree" to be more descriptive of what you get back
from this macro.

But I'm happy to close this without merging if others are attached to "deep."

EDIT: maybe the "deep" here is in reference to `deepcopy`? I hadn't actually made that connection, so maybe the original is fine as is. Let's see if others chime in with thoughts on the matter.

### DIFF
--- a/SnoopCompileCore/src/SnoopCompileCore.jl
+++ b/SnoopCompileCore/src/SnoopCompileCore.jl
@@ -12,7 +12,7 @@ if VERSION >= v"1.2.0-DEV.573"
 end
 
 if VERSION >= v"1.6.0-DEV.1190"  # https://github.com/JuliaLang/julia/pull/37749
-    include("snoopi_deep.jl")
+    include("snoopi_tree.jl")
 end
 
 if VERSION >= v"1.6.0-DEV.154"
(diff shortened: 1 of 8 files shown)


timholy:
I don't think @sacha0 is watching this, but a change like this should get some eyeballs on it. https://github.com/timholy/SnoopCompile.jl/pulls?q=is%3Apr+is%3Aclosed

Sacha0:
👍 more descriptive! :)

timholy:
Sounds good. I'm in the middle of finishing off a more comprehensive version of #180; I'll resubmit this after that merges. (There will be a lot of conflicts.)

Sacha0:
Cheers, please feel welcome to ping me for review and I'll do my best to work it in (though regrettably likely with relatively high latency, and after finishing off https://github.com/JuliaLang/www.julialang.org/pull/1093).

Hope you have been having and continue to have a lovely holiday season! :)

timholy:
That version of the blog post will have to be pretty extensively rewritten. I've iterated several times between "use it in anger" and "polish the rough edges," and there have been a lot of changes; not just to the API, but how I think about tackling this. (The seeds are all there, but it is still going to be a pretty big rewrite.)

So I wouldn't yet put much time into that post. My main thought is to now split it into a series of posts; the first will have background material, and is now at https://github.com/JuliaLang/www.julialang.org/pull/1111. Feel free, or not, to review it.

timholy:
I'm going to just stick with `@snoopi_deep`.

PR #191, opened 2020-12-29
snoopi_deep: return inclusive & exclusive times

timholy:
This modifies `@snoopi_deep` to return "normalized" data that allows all analyses to have access to both inclusive and exclusive time. It also converts all times to `secs::Float64`. Finally, it does a full refresh on all the docstrings to account for the changes in results and their display.
    
Supersedes and closes #180. Other than #190 and the documenter docs, this is the last change I expect before the big release.

CC @Sacha0.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,7 +25,6 @@ YAML = "0.4"
 julia = "1"
 
 [extras]
-AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 ColorTypes = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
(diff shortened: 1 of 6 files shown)


Sacha0:
Added to the review queue; let me know if you have a particular timescale in mind for merge! :)

timholy:
I can wait a few days, but realistically I need to clear this from my todo list soon so I can focus on other things. If you don't get time to review it, no real worries, though I have not a shred of a doubt that it would improve due to your review.

The main thing I'm concerned about is breaking things in ways that will make you folks unhappy. You two helped get this whole thing going, but I've made quite a few changes. To me they all feel like improvements, but :man_shrugging:

timholy:
See also #191,which if you scroll down to the documenter/deploy workflow it will take you to a preview of the documentation.

PR #207, opened 2021-01-07
Fix a flaky runtime/inferencetime test

timholy:
There was really no solid justification for the previous test. This new
version is based on more solid footing.


timholy:
OK, the real problem seems to have been that a lot of the costs formerly were attributed to C frames, and it seems the balance of costs was either fairly random or somewhat platform-dependent. This now attributes all cost to Julia frames, and preliminarily seems much more reliable.

While we were analyzing these frames in greater detail, it seemed like a good chance to incorporate runtime dispatch, and so this also stores this information. Technically the change in output format is breaking, but this tool wasn't documented yet, so let's not worry about it.

PR #208, opened 2021-01-08
README.md: update build status badge

fingolfin:
Point to GitHub Actions instead of Travis.


timholy:
Thanks so much! I always forget about the badges.

PR #212, opened 2021-01-11
A grabbag of random fixes

timholy:
These are things I fixed while using this in practice. Unfortunately I don't have tests for them (bad, bad), I guess when I encountered them in practice I felt it was important to keep moving.

Better to have them than not, I think.

### DIFF
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -911,8 +911,8 @@ The empty horizontal periods in the flamegraph correspond to times when somethin
 The total width of the flamegraph is set from the `ROOT` node.
 """
 function FlameGraphs.flamegraph(tinf::InferenceTimingNode; tmin = 0.0, excluded_modules=Set([Main::Module]), mode=nothing)
+    isROOT(tinf) && isempty(tinf.children) && error("root node has no children")
     io = IOBuffer()
-
     # Compute a "root" frame for the top-level node, to cover the whole profile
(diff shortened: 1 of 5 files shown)


src/parcel_snoopi_deep.jl
```diff
@@ -905,8 +911,8 @@ The empty horizontal periods in the flamegraph correspond to times when somethin
 The total width of the flamegraph is set from the `ROOT` node.
 """
 function FlameGraphs.flamegraph(tinf::InferenceTimingNode; tmin = 0.0, excluded_modules=Set([Main::Module]), mode=nothing)
+    isROOT(tinf) && isempty(tinf.children) && error("root node has no children")
```
NHDaly, line 914: @timholy I just hit this error in usage of this package..

It seems to me that it's much better to just return an otherwise empty profile, with just the single ROOT node.
From what i can tell, without this check, it _does_ return that empty profile correctly.

Can we remove this line? I'm curious if you remember the motivation to add it.

timholy: Very happy to have you change this though!

NHDaly: Yeah, makes sense. I think i was thinking that if it's empty, then it's empty! 😅  But you're right that it's not as nice as if it said "hey this is empty."

Maybe we can print a warning instead of throwing an error?

Our use case is that we're creating this via an HTTP endpoint on prod servers now (!!), and i don't want that to throw an error and return a 500:
https://github.com/JuliaPerf/ProfileEndpoints.jl/pull/16

K, thanks, i'll send a quick PR

PR #240, opened 2021-03-03
Remind people to use startup-file=no

oxinabox:
I keep forgetting to do this.


timholy:
Looks like precompilation is broken on nightly for windows & osx.

Thanks for the improvement!

PR #245, opened 2021-07-12
Add `@snoopl` to reference docs

NHDaly:
I just noticed that `@snoopl` isn't in the documentation at all. I know it's not very widely used, but it'd be nice to refer to it here at least. :)

`@snoopl` was originally added in https://github.com/timholy/SnoopCompile.jl/pull/135/.

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -7,6 +7,7 @@
 @snoopi_deep
 @snoopi
 @snoopc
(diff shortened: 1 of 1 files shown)


timholy (approved):
Seems fine to me! Some more documentation might increase its utility, but TBH I haven't played with it enough to know.

Issue #251, opened 2021-07-24
inference/LLVM codegen/native codegen time split


bkamins:
If I understand things correctly SnoopCompile.jl currently allows to dig into inference timings. It would be great to be able to similarly dig into LLVM codegen and native codegen timings.

This is relevant for packages like DataFrames.jl, where in many cases it is LLVM codegen and/or native codegen timing a chief time of running some functions (and this is unavoidable since for many functions users pass anonymous functions as an argument).

An example of such a case is here: https://github.com/JuliaData/DataFrames.jl/issues/2806

NHDaly:
@bkamins you already currently can dig into LLVM optimization and native codegen timings, via `@snoopl` and `@snoopc`.

We haven't written big blog posts about those yet, I think, but there is at least some information in the documentation site:
- Snoopc: https://timholy.github.io/SnoopCompile.jl/stable/snoopc/
- Snoopl currently only has these two docstrings:
    - https://timholy.github.io/SnoopCompile.jl/stable/reference/#SnoopCompileCore.@snoopl
    - https://timholy.github.io/SnoopCompile.jl/stable/reference/#SnoopCompile.read_snoopl


Part of the difficultly, though, is that the signal you get from those two measurements is less useful than the signal you get from `@snoopi_deep`.

This is because for inference, julia currently performs type inference via a depth-first traversal, so by measuring trace spans, we get a nice flamegraph that you can look at, to understand both inclusive and exclusive inference timings.

For LLVM optimization on the other hand, we pass functions through LLVM optimization in a somewhat random order, batched together arbitrarily, which makes it more difficult to account optimization time to individual functions. `@snoopl` currently emits:
1. An array of Pairs of: the amount of time compiling each "batch", or `module`, and the type signature for the method instances in that module.
2. A Dict of before/after statistics about each method instance in the module, mapping type signature to stats like number of instructions before/after optimization.

And finally, I believe `@snoopc` provides a mapping of the time to codegen each method instance.

--------

Hope that's helpful! :) If you find any of of that useful, documentation PRs are always welcome ❤️ :)

bkamins:
> you already currently can dig into LLVM optimization and native codegen timings, via `@snoopl` and `@snoopc`.

The problem is that they start a new Julia session and I need to run them in an already running session so that I do not measure the cost of first compilation. I am interested only in consecutive compilation times. Is this possible?

bkamins:
Reproducible example on Julia nightly:
```
               _
[...]

PR #252, opened 2021-07-25
Add option to snoop commands to control spawning separate process vs eval'ing in existing process

NHDaly:
Running snoop compilation in a separate process allows you to consistently measure complete compilation time on every run, without being affected by the state of the current process.

On the other hand, running the snoop compilation in the current process allows you to measure _incremental_ compilation performance, which can also be useful in some contexts.

This PR attempts to add an option to all snoop commands to let you choose between those options.
It also adds another function, `@snoop_all`, which runs `@snoopi`, `@snoopl` and `@snoopc` together at the same time, so that you can compare results across them.

-----------------------

Note that this branch is currently really messy, because it had a bunch of draft PRs merged into it, that have since been squash-merged into master, so we need to clean that up first as well.

This addresses this comment, for a request from @bkamins: https://github.com/timholy/SnoopCompile.jl/issues/251#issuecomment-886093390


bkamins:
Thank you. Can you please ping me when this is ready to test? (or if you want me to review it now I can also try but I do not know the code base much)

SnoopCompileCore/src/SnoopCompileCore.jl
```diff
@@ -23,4 +23,8 @@ if VERSION >= v"1.6.0-DEV.1192"  # https://github.com/JuliaLang/julia/pull/37136
     include("snoopl.jl")
 end
 
+if VERSION >= v"1.6.0-DEV.1192"  # https://github.com/JuliaLang/julia/pull/37136
+    include("snoop_all.jl")
```
timholy, line 27: Can't we just include this macro into `SnoopCompileCore/src/snoopl.jl`? It's the same condition for including it that we use for`@snoopl`.

timholy (commented):
LGTM, just one small comment.

A test would be nice, of course!

SnoopCompileCore/src/SnoopCompileCore.jl
NHDaly: Yeah, good idea. I've just moved it into the same if-block. I guess it doesn't necessarily belong in `snoopl.jl` because it will snoop all three types of compilation, which is why I put it in its own file.

Can you have another look and let me know if this looks okay? :)

NHDaly:
So i think this is more or less ready to go, but for some reason that I can't explain, the first call to `@snoopl` is now failing with an `IOError: write: broken pipe (EPIPE)` during the `serialize()`. But only the first time, when i run it in the REPL, and then it seems to succeed consistently after that.

I have no idea why. In the last commit, 995ffd2, I tried simplifying the logic that's spawning the processes and passing the commands through, but it had no effect.

This is also why the tests are failing.

Any advice you have would be greatly appreciated!! 😬 🤞 

=================

Otherwise, I've added tests and a docstring.

There's still no standalone tests for `@snoopc` and `@snoopl` with `pspawn=false`, and it would be nice to add those.

Finally, there's still no `pspawn=` option for `@snoopi_deep` or `@snoop_all`, which we could consider adding to make them more like the other functions. But this can be done in a follow-up PR.

NHDaly:
Tl;dr: Tests failing and i don't know why - plz help!! 😬

NHDaly:
@bkamins - you could try testing this now if you'd like

bkamins:
Thank you. I do not get the error you report. However I have some general minor comments:
* running this macro swamps my terminal with the output; is this intended?
* the two files you give .csv extension are not comma separated (it seems they are tab separated)
* could you please guide me what do the numbers in `*snoopc.csv` file mean (these in the beginning of each line)

Thank you!

bkamins:
Apart from that I have analyzed the results - and all works as I would expect. Superb work - thank you!

@pdeffebach - given the results of `@snoop_all` the difference between `subset` and `select` in compilation time is unfortunately due to the additional compilation that `subset` has (the extra logic it has). This is probably unavoidable (and this was expected).

NHDaly:
> * running this macro swamps my terminal with the output; is this intended?

I've added an example with a `;` at the end.
Probably we should change the default show for the `InferenceTimingNode` returned as the first argument (`tinf`) from `@snoopi_deep` so that it doesn't print such a big tree, as well..

> the two files you give .csv extension are not comma separated (it seems they are tab separated)
Huh, until today I thought that you could have arbitrary delimiters, but it's still called a CSV.  Should we rename them to `.tsv`?
https://superuser.com/a/577912

> could you please guide me what do the numbers in *snoopc.csv file mean (these in the beginning of each line)

Please see the updated docstring, with links to what to do with each of the results. You should be able to follow all of those separate guides for how to interpret the results. :)

bkamins:
Thank you!

SnoopCompileCore/src/snoopc.jl
```diff
@@ -4,33 +4,50 @@ using Serialization
...
+function snoopc(flags, filename, commands, _module=nothing; pspawn=true)
+    if pspawn
+        println("Launching new julia process to run commands...")
+    end
     code_object = """
             using Serialization
-            while !eof(stdin)
-                Core.eval(Main, deserialize(stdin))
-            end
+            Core.eval(Main, deserialize(IOBuffer(read(stdin))))
```
timholy, line 48: Any particular reason for this change? What if `stdin` is *huge*? Likewise elsewhere among the changes.

Is it possible that this is the change that causes the closed stream issue? It seems possible that Julia 1.0 is not being as careful as this code.

timholy (commented):
Looks quite good. For consistency, let's re-export `@snoop_all` from SnoopCompile.

SnoopCompileCore/src/snoopc.jl
NHDaly: Alas, this is not responsible for the error -- this was my attempt to fix it, but it didn't help. I can happily revert these changes, yeah.

kimikage:
This is not a technical issue, but the name `all` seems somewhat misleading.
Perhaps this PR needs to be targeted and updated for SnoopCompile v3 or v2, but there is room for reconsideration of the name as well.

NHDaly:
> This is not a technical issue, but the name `all` seems somewhat misleading.

How so? What would you prefer instead? The idea was that it should perform all the analyses we have: inference, llvm, etc.
I'm open to something better. Perhaps `@snoop_all_phases`?

NHDaly:
Also, sorry I let this PR completely rot. 😢

PR #254, opened 2021-08-11
minor documentation improvements

aviatesk:
While reading through the documentation, I found some words might look
nicer if they are codified (i.e. by enclosing them by backticks), mainly:
- make sure to codify `MethodInstance`
- if a package module is actually used, codify its name, e.g. `SnoopCompile`

Please tell me if you don't like these changes -- I don't mind simply
reverting that part.

Unrelated to above changes, I also added some typo fixes.


timholy (approved):
Thanks, this is great! Merge at will :slightly_smiling_face:

PR #260, opened 2021-08-18
Track invalidations during deserialization

timholy:
These were printed to the console but not otherwise stored.
Storing them allows one to give them more emphasis and perform
analysis. With https://github.com/JuliaLang/julia/pull/41913,
it becomes possible to attribute them to specific method
definitions or deletions. (Of course there might be multiple
reasons, but we need to identify at least one in order to make
progress.)

CC @chriselrod, @vtjnash, @ChrisRackauckas. If anyone plans to try this, until 2.8 is released be sure to `dev SnoopCompileCore` too as it is a separate (subdir) module and there are changes due to #257.

### DIFF
--- a/src/invalidations.jl
+++ b/src/invalidations.jl
@@ -74,6 +74,10 @@ mutable struct InstanceNode
     end
 end
 
+Core.MethodInstance(node::InstanceNode) = node.mi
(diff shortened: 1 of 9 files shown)


timholy:
To really be functional in the tests we'll have to wait for JuliaLang/julia#41913 to hit nightly. But since this hasn't broken anything, I'll merge.

PR #262, opened 2021-08-19
Identify invalidating blockers of precompilation

timholy:
This allows one to identify the particular invalidations that block
precompilation of particular MethodInstances.

I am not 100% sure this is right. In particular, it might be too conservative; maybe one should `SnoopCompile.getrroot(hits[i])` to see the real impact? But I checked whether the root CodeInstances are there in the precompile cache, and they seem to be. So my best guess is that this is painting an accurate picture as-is. I might let this sit on master for a few days before making a release. 

CC @ChrisRackauckas. It seems to suggest that invalidation is not your first-order problem with https://github.com/SciML/ModelingToolkit.jl/pull/1215, as you seemed to expect. But I wanted to get this tool out there in some form.

### DIFF
--- a/src/SnoopCompile.jl
+++ b/src/SnoopCompile.jl
@@ -106,6 +106,11 @@ if isdefined(SnoopCompileCore, Symbol("@snoopr"))
     export @snoopr, uinvalidated, invalidation_trees, filtermod, findcaller, ascend
 end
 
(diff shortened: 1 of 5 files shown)


timholy:
I should say that for this to work well you need to be running at least Julia 1.8.0-DEV.368 (you need https://github.com/JuliaLang/julia/pull/41913)

PR #268, opened 2021-12-05
Fix string cat in show method invalidations

charleskawczynski:
This PR fixes a small typo for concatenating strings in the `show` method for invalidations.

Thanks for the great package!


timholy:
Thanks! Do you have a reproducer for this? I generally attempt to capture the problem in a test before making the fix, to ensure no similar problems come up in the future.

charleskawczynski:
I don't, but I can try to cook one up

timholy:
Don't kill yourself for it, because it is indeed an obvious fix, but I thought I'd ask.

charleskawczynski:
This is as far as I've gotten:
```julia
using SnoopCompileCore

invalidations = @snoopr begin
    import BlockArrays
    import CUDA
end;

using SnoopCompile
trees = invalidation_trees(invalidations)
```

timholy:
Thanks for the tip! I added a test that doesn't require those big packages. Here's how I captured the error: with SnoopCompile `dev`ved, I identified method that threw the error from the stacktrace and added this (Revise.jl makes this easy):
```diff
diff --git a/src/invalidations.jl b/src/invalidations.jl
index f8fd2db..37df189 100644
--- a/src/invalidations.jl
+++ b/src/invalidations.jl
@@ -173,7 +173,9 @@ end
 
 # We could use AbstractTrees here, but typically one is not interested in the full tree,
 # just the top method and the number of children it has
+const minvs = Ref{MethodInvalidations}()
 function Base.show(io::IO, methinvs::MethodInvalidations)
+    minvs[] = methinvs
     iscompact = get(io, :compact, false)::Bool
 
     print(io, methinvs.reason, " ")
```
This allowed me to extract (in the REPL0 the most recent input to the function via `minvs = SnoopCompile.minvs[]`. Then I was able to inspect it, identify which method definition in CUDA was responsible for the invalidation, and just mimic that.

PR #272, opened 2021-12-08
Color constprop-nodes in flamegraph

timholy:
Constant-propagation can be an unexpected source of recompilation,
so it makes sense to highlight it. This piggy-backs off of FlameGraph's
marking of GC-events (default coloration yellow) to mark nodes
where one or more non-function arguments has constant-propagation.

TODO:
- [x] update docs


timholy:
This was used to identify targets for https://github.com/JuliaLang/julia/pull/43367

NHDaly (commented):
Cool change, BTW! :) 👍

PR #269, opened 2021-12-06
Skip nothing results in visualization to avoid log(::Nothing)

charleskawczynski:
I've been visualizing invalidations for a package, the plots and docs are fantastic. I have been running into an error when plotting, however:

```
RuntimeError: <PyCall.jlwrap (in a Julia function called from Python)
JULIA: MethodError: no method matching log(::Nothing)
Closest candidates are:
  log(!Matched::StridedMatrix{T} where T) at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/LinearAlgebra/src/dense.jl:757
  log(!Matched::SymbolicUtils.Add) at ~/.julia/packages/SymbolicUtils/fgHzN/src/methods.jl:68
  log(!Matched::SymbolicUtils.Sym) at ~/.julia/packages/SymbolicUtils/fgHzN/src/methods.jl:68
  ...
Stacktrace:
  [1] (::SnoopCompile.var"#onclick#171"{Float64, Base.RefValue{Union{Method, SnoopCompile.MethodLoc}}})(event::PyCall.PyObject)
    @ SnoopCompile ~/.julia/packages/SnoopCompile/NF9Dd/src/visualizations.jl:32
  [2] invokelatest(::Any, ::Any, ::Vararg{Any}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Base ./essentials.jl:716
  [3] invokelatest(::Any, ::Any, ::Vararg{Any})
    @ Base ./essentials.jl:714
  [4] _pyjlwrap_call(f::Function, args_::Ptr{PyCall.PyObject_struct}, kw_::Ptr{PyCall.PyObject_struct})
    @ PyCall ~/.julia/packages/PyCall/3fwVL/src/callback.jl:32
  [5] pyjlwrap_call(self_::Ptr{PyCall.PyObject_struct}, args_::Ptr{PyCall.PyObject_struct}, kw_::Ptr{PyCall.PyObject_struct})
    @ PyCall ~/.julia/packages/PyCall/3fwVL/src/callback.jl:44
  [6] macro expansion
    @ ~/.julia/packages/PyCall/3fwVL/src/exception.jl:95 [inlined]
  [7] #107
    @ ~/.julia/packages/PyCall/3fwVL/src/pyfncall.jl:43 [inlined]
  [8] disable_sigint
    @ ./c.jl:458 [inlined]
  [9] __pycall!
[...]

PR #273, opened 2022-01-22
Detect stray clicks in pgdsgui

timholy:
Fixes #269


timholy:
Test failure on nightly is #274

PR #281, opened 2022-07-05
make SnoopCompile compatible with JET@0.6.0

aviatesk:


aviatesk:
The failures in nightly seems to come from the changes on Julia Base from https://github.com/JuliaLang/julia/pull/45069. The integration with JET runs successfully.

timholy:
Thanks so much!

PR #282, opened 2022-07-19
Add SnoopPrecompile package

timholy:
This small package makes it easy to exhaustively precompile on
Julia 1.8 and higher. In brief, your package should look like

```julia
module MyPackage

using SnoopPrecompile

# method definitions etc

let
    @precompile_calls begin
        # execute the code you want precompiled
    end
end

end # module MyPackage
```

There are also facilities for "setup" code, see the documentation
for details.

### DIFF
--- a/SnoopPrecompile/Project.toml
+++ b/SnoopPrecompile/Project.toml
@@ -8,6 +8,7 @@ julia = "1"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [targets]
-test = ["Test"]
+test = ["Test", "UUIDs"]
--- a/SnoopPrecompile/Project.toml
+++ b/SnoopPrecompile/Project.toml
@@ -1,0 +1,13 @@
+name = "SnoopPrecompile"
+uuid = "66db9d55-30c0-4569-8b51-7e840670fc0c"
+authors = ["Tim Holy <tim.holy@gmail.com> and contributors"]
+version = "1.0.0"
+
+[compat]
+julia = "1"
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["Test"]
--- a/docs/src/snoop_pc.md
+++ b/docs/src/snoop_pc.md
@@ -20,12 +20,10 @@ struct OtherType
     str::String
 end
 
-let  # `let` prevents `list` from being a visible private name in the package
-    @precompile_calls :setup beginNaturally, y
-        # Putting some things in `:setup` can reduce the size of the
-        # precompile file and potentially make loading faster.
-        list = [OtherType("hello"), OtherType("world!")]
-    end
+@precompile_calls :setup begin
+    # Putting some things in `:setup` can reduce the size of the
+    # precompile file and potentially make loading faster.
+    list = [OtherType("hello"), OtherType("world!")]
     @precompile_calls begin
         # all calls in this block will be precompiled, regardless of whether
         # they belong to your package or not (on Julia 1.8 and higher)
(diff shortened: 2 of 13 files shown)


timholy:
CC @ChrisRackauckas, @SimonDanisch, @rikhuijzer, @LilithHafner, @IanButterworth. This gets both the inferrable and runtime-dispatched calls.

rikhuijzer:
So basically this approach falls in between running a workload and calling `precompile` manually? Compared to the workload, it doesn't cause side-effects and compared to calling `precompile` it will also call runtime dispatches?

Why is `SnoopPrecompile` located inside the `SnoopCompile` repo and not in a separate one?

LilithHafner:
>
```julia
let
    @precompile_calls begin
        # execute the code you want precompiled
    end
end
```

Wouldn't it be easier for the macro to introduce the let block?

SnoopPrecompile/Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "SnoopPrecompile"
+uuid = "66db9d55-30c0-4569-8b51-7e840670fc0c"
+authors = ["Tim Holy <tim.holy@gmail.com> and contributors"]
+version = "1.0.0"
+
+[compat]
+julia = "1"
```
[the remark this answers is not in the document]
timholy: I think of this as a "support" package (one that does not really exist in its own right but merely to make others better), and this allows it to support packages that happen to still support Julia 1.0. It won't do much useful on 1.0, but this way it doesn't force anyone to abandon 1.0 support.

timholy:
> Wouldn't it be easier for the macro to introduce the let block?

Handling `:setup` code that way would have required
```julia
@precompile_calls :setup begin
    vars = ...
    @precompile calls begin
        somecall(vars...)
    end
end
```
since `vars` needs to be in scope for `somecall(vars...)`. I got the impression you didn't like the nested approach. But I can go that way if you prefer it.

LilithHafner:
Using that approach, `@precompile_calls` without `:setup` could still let-fence itself.

Another approach uses only a single macro call, following the style of `BenchmarkTools.@benchmark`:
```julia
@precompile_calls begin
    somecalls(vars...)
end setup = (vars = ...)
```

timholy:
> For the macro name @precompile_calls, I would suggest starting with a verb since most functions and macro's either start with a verb (e.g., @time, @benchmark and @debug) or state the output (e.g., @code_llvm and @lazy_str). Maybe @call_precompile?

It does :slightly_smiling_face: . I'm `precompile`ing the `calls` in that expression!

timholy:
But of course happy to accept other names, given that it's obviously confusing!

LilithHafner:
> It does 🙂 . I'm `precompile`ing the `calls` in that expression!

That is 100% obvious now that you say it, but I also didn't realize that precompile was a verb in that context.

docs/src/snoop_pc.md
```diff
@@ -0,0 +1,70 @@
...
+
+struct MyType
+    x::Int
+end
+struct OtherType
+    str::String
+end
+
+let  # `let` prevents `list` from being a visible private name in the package
+    @precompile_calls :setup beginNaturally, y
```
jkrumbiegel, line 24: What is `beginNaturally, y` here?

timholy: Weird vscode paste! Should just be `begin`. Every once in a while vscode does weird things to my text, I must be sloppy about exiting multicursor or something?

timholy:
If a name change doesn't seem in the offing, clearly it needs at least a docstring.

timholy:
I think I've addressed most of the feedback. Some comments:

- I was worried that there could be a large amount of setup code, so the BenchmarkTools `setup = (...)` syntax began to seem a little less appealing. I went with the nested approach. Alternatively we could have `@precompile_setup` and `@precompile_calls`.
- Sadly, we seem to have a bug in our ability to disable the interpreter, so automatic `let`-fencing is not yet possible. Until then, you can use manual fencing at top level, but be sure to check results with `verbose[] = true`.

timholy:
I'm beginning to lean towards `@precompile_setup` and `@precompile_all`, since people found `@precompile_calls` confusing. Thoughts?

rikhuijzer:
+1 for `@precompile_all`

jkrumbiegel:
How about `@precompile_all_calls`, it's longer but clear in that calls are the object not the verb, and otherwise you don't know from reading what's `all`.

timholy:
I went with `@precompile_setup` and `@precompile_all_calls`. Barring any additional comments I will merge and register the package.

timholy:
https://github.com/JuliaRegistries/General/pull/64746

PR #284, opened 2022-07-30
fix typo in SnoopPrecompile docs

alecloudenback:
I think `all_begin` was intended to be `begin`? 

The former errorred for me when I tried to add it to a package.


timholy:
Thanks!

PR #297, opened 2022-09-20
Fix warning during precompilation

thchr:
https://github.com/JuliaLang/julia/pull/46608 gave rise to a warning during precompilation of SnoopCompile:

```jl
julia> using SnoopCompile
[ Info: Precompiling SnoopCompile [aa65fe97-06da-5843-b5b1-d5d13cad87d2]
WARNING: method definition for precompilable_time at /home/tchr/.julia/packages/SnoopCompile/J3dtR/src/parcel_snoopi_deep.jl:290 declares type variable T but does not use it.
```


timholy:
Thanks!

Issue #298, opened 2022-09-21
Option to turn off all precompilation globally


baggepinnen:
Since the introduction of SnoopPrecompile, I've noticed a rather dramatic impact on my developer productivity. As a developer of many different packages, many of them related to SciML, I precompile packages very often, and have very little benefit of any precompilation, which has lately been wasting 20 minutes multiple times per day.

I just learned that OrdinaryDiffEq is using Preferences.jl to allow the user to opt out of precompilation. Would it be possible to do something similar for SnoopPrecompile globally, using Preferences.jl or an environment variable? I would essentially like to turn off more or less *all* precompilation, since chances are rather high that I will not benefit from it at all until the next time I perform the same precompilation.

timholy:
What happens if you start julia with `julia --compiled-modules=no`?

baggepinnen:
`--compiled-modules=no` seems to make loading packages extremely slow
```julia
/home/fredrikb> julia --startup-file=no --compiled-modules=no 

julia> @time using ControlSystems
┌ Warning: Replacing docs for `SciMLBase.sol :: Union{Tuple, Tuple{D}, Tuple{S}, Tuple{N}, Tuple{T}} where {T, N, S, D}` in module `SciMLBase`
└ @ Base.Docs docs/Docs.jl:240
 81.987430 seconds (132.87 M allocations: 8.530 GiB, 3.21% gc time, 34.21% compilation time: 9% of which was recompilation)
```

```julia
/home/fredrikb> julia --startup-file=no --compiled-modules=yes

julia> @time using ControlSystems
[ Info: Precompiling ControlSystems [a6e380b2-a6ca-5380-bf3e-84a91bcd477e]
 15.364393 seconds (19.27 M allocations: 1.285 GiB, 2.41% gc time, 4.63% compilation time: 48% of which was recompilation)
```

timholy:
Hmm, yeah, it also nixes the caching of lowered code.

PR #299, opened 2022-09-21
Add option to globally disable snoop precompilation

baggepinnen:
This implements one of the suggestions from #298 and allows the user to opt out of all precompilation by setting the environment variable `ENV["SNOOP_PRECOMPILE"] = false`

### DIFF
--- a/SnoopPrecompile/src/SnoopPrecompile.jl
+++ b/SnoopPrecompile/src/SnoopPrecompile.jl
@@ -44,6 +44,7 @@ end
         - indirect runtime-dispatched calls to such methods.
 """
 macro precompile_all_calls(ex::Expr)
+    get(ENV, "SNOOP_PRECOMPILE", true) || return :()
     if have_force_compile
         ex = quote
             begin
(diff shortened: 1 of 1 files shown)


timholy:
The issue I see with this is that it may leave people in a complicated mix of `*.ji` files: those that happened to be written while `SNOOP_PRECOMPILE=0` and those written when it defaulted to being on. That means TTFX could vary wildly depending on the specifics of your precompilation history. Morever, this doesn't seem to have a mechanism for ensuring that the precompiled versions get generated & used when you stop setting `SNOOP_PRECOMPILE=0`. The only option for a "clean house" would be to delete all the files in your `.julia/compiled/v1.x` directory and then precompile everything.

However, I think if we use Preferences.jl, all these negatives go away. Care to try an implementation using Preferences and see if it solves your problem?

timholy:
Just checking in, is this something you're planning to do or should I marked this "help wanted"?

baggepinnen:
I'd be happy to do this, I've been out-of-office for a while but will try to get to it after the backlog has been worked through.

timholy:
With native-code compilation about to be merged, this will become quite urgent. I'll tackle it myself within a week unless you tell me otherwise.

timholy:
I think we also want the ability to turn it off for specific packages, again under `Preferences.jl` control.

t-bltg:
Superseded by https://github.com/timholy/SnoopCompile.jl/pull/334.

PR #300, opened 2022-09-24
Integrate JET via Requires.jl

timholy:
Both SnoopCompile and JET are difficult to keep compatible with nightly and across older versions of Julia, so requiring that both are working (or even, just precompiling) simultaneously is twice the risk. If we put the JET integration under `@require`, then we can load SnoopCompile without JET.

The sad part is that this should probably be viewed as a breaking change (now users have to manually load JET) and thus we need to go to 3.0.0.


mkitti:
This is made obsolete by https://github.com/timholy/SnoopCompile.jl/pull/379

aviatesk:
Yeah, this is superseded by #379.

PR #303, opened 2022-09-30
Add `report_invalidations` to tabulate invalidations summary

charleskawczynski:
Hello 👋🏻! There has been a lot of activity around hunting down invalidations lately, for example:

 - [julia-invalidations](https://github.com/julia-actions/julia-invalidations)
 - [discourse thread](https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359)
 - [SciML issue](https://github.com/SciML/DifferentialEquations.jl/issues/786)

As mentioned in [this comment](https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359/21) of the `discourse thread`, I've added a function called `report_invalidations` to [ReportMetrics.jl](https://github.com/CliMA/ReportMetrics.jl) which summarizes the output of `SnoopCompileCore.@snoopr` with quantitative and actionable information.

I'd like to call this function from the `julia-invalidations` github action. I thought it might be cleaner to add this functionality to SnoopCompile directly (rather than a separate package) since it's relatively light weight (barring PrettyTables) and uses SnoopCompile internals.

The `ReportMetrics.jl` version of the printed table looks like this:
```
[ Info: Number of invalidations for SomeWork: 7
[ Info: SomeWork: 6 total method invalidations
┌───────────────────────────────────┬─────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>         │ Method Name │ Invalidations │ Invalidations % │
│                                   │             │    Number     │     (xᵢ/∑x)     │
├───────────────────────────────────┼─────────────┼───────────────┼─────────────────┤
│ SnoopCompile.jlirrationals.jl:182 │  BigFloat   │       1       │       50        │
│ SnoopCompile.jlirrationals.jl:182 │  BigFloat   │       1       │       50        │
│ SnoopCompile.jlirrationals.jl:192 │   Float32   │       0       │        0        │
│ SnoopCompile.jlirrationals.jl:191 │   Float64   │       0       │        0        │
│ SnoopCompile.jlirrationals.jl:192 │   Float32   │       0       │        0        │
│ SnoopCompile.jlirrationals.jl:191 │   Float64   │       0       │        0        │
└───────────────────────────────────┴─────────────┴───────────────┴─────────────────┘
```

### DIFF
--- a/src/invalidations.jl
+++ b/src/invalidations.jl
@@ -256,8 +256,8 @@ else
 end
 
 """
-    report_invalidations(;
-        job_name::String = "",
+    report_invalidations(
+        io::IO = stdout;
         invalidations,
         n_rows::Int = 10,
         process_filename::Function = x -> x,
@@ -265,10 +265,19 @@ end
 
 Print a tabular summary of invalidations given:
 
- - `job_name::String` the name of the job
+ - `invalidations` the output of [`SnoopCompileCore.@snoopr`](@ref)
 
-An `@info` message also prints some information, for example
-if the table has been truncated (upon request).
+and (optionally)
+
+ - `io::IO` IO stream. Defaults to `stdout`
+ - `n_rows::Int` the number of rows to be displayed in the
+   truncated table. A value of 0 indicates no truncation.
+   A positive value will truncate the table to the specified
+   number of rows.
+ - `process_filename(::String)::String` a function to post-process
+   each filename, where invalidations are found
+
+# Example usage
 
 ```julia
 import SnoopCompileCore
@@ -280,10 +289,7 @@ end;
 
 using SnoopCompile
 using PrettyTables # to load report_invalidations
-report_invalidations(;
-    job_name = "MyWork",
-    invalidations,
-)
+report_invalidations(;invalidations)
 ```
 
 Using `report_invalidations` requires that you first load the `PrettyTables.jl` package.
--- a/src/invalidations.jl
+++ b/src/invalidations.jl
@@ -256,16 +256,15 @@ else
 end
 
 """
-    tabulated_invalidations(;
+    report_invalidations(;
         job_name::String,
         invalidations,
         n_rows::Int = 10,
         process_filename::Function = x -> x,
     )
 
-A NamedTuple, `(; table_data, header)`, which
-contains a condensed form of invalidations in the form
-of a matrix and a header, which can be easily tabulated.
+Print a tabular summary of invalidations.
+
 An `@info` message also prints some information, for example
 if the table has been truncated (upon request).
 
@@ -277,66 +276,16 @@ invalidations = SnoopCompileCore.@snoopr begin
 
 end;
 
-import SnoopCompile
-(; table_data, header) = SnoopCompile.tabulated_invalidations(;
+using SnoopCompile
+report_invalidations(;
     job_name = "MyWork",
     invalidations,
 )
-
-import PrettyTables
-PrettyTables.pretty_table(
-    table_data;
-    header,
-    formatters = PrettyTables.ft_printf("%s", 2:2),
-    header_crayon = PrettyTables.crayon"yellow bold",
-    subheader_crayon = PrettyTables.crayon"green bold",
-    crop = :none,
-    alignment = [:l, :c, :c, :c],
-)
 ```
+
+Using `report_invalidations` requires that you first load the `PrettyTables.jl` package.
 """
-function tabulated_invalidations(;
-        job_name::String,
-        invalidations,
-        n_rows::Int = 10,
-        process_filename::Function = x -> x,
-    )
-    trees = reverse(invalidation_trees(invalidations))
-    n_total_invalidations = length(uinvalidated(invalidations))
-    # TODO: Merge `@info` statement with one below
-    @info "Number of invalidations for $job_name: $n_total_invalidations"
-    invs_per_method = map(trees) do methinvs
-        countchildren(methinvs)
-    end
-    n_invs_total = length(invs_per_method)
-    truncated_invs = n_rows < n_invs_total
-    sum_invs = sum(invs_per_method)
-    invs_per_method = invs_per_method[1:min(n_rows, n_invs_total)]
-    trees = trees[1:min(n_rows, n_invs_total)]
-    trunc_msg = truncated_invs ? " (showing $n_rows) " : ""
-    @info "$(job_name): $n_invs_total total method invalidations$trunc_msg"
-    n_invalidations_percent = map(invs_per_method) do inv
-        inv_perc = inv / sum_invs
-        Int(round(inv_perc*100, digits = 0))
-    end
-    meth_name = map(trees) do inv
-        "$(inv.method.name)"
-    end
-    fileinfo = map(trees) do inv
-        "$(process_filename(string(inv.method.file))):$(inv.method.line)"
-    end
-    header = (
-        ["<file name>:<line number>", "Method Name", "Invalidations", "Invalidations %"],
-        ["", "", "Number", "(xᵢ/∑x)"],
-    )
-    table_data = hcat(
-        fileinfo,
-        meth_name,
-        invs_per_method,
-        n_invalidations_percent,
-    )
-    return (; table_data, header)
-end
+function report_invalidations end
 
 """
     trees = invalidation_trees(list)
--- a/src/invalidations.jl
+++ b/src/invalidations.jl
@@ -256,6 +256,89 @@ else
 end
 
 """
+    tabulated_invalidations(;
+        job_name::String,
+        invalidations,
+        n_rows::Int = 10,
+        process_filename::Function = x -> x,
+    )
+
+A NamedTuple, `(; table_data, header)`, which
+contains a condensed form of invalidations in the form
+of a matrix and a header, which can be easily tabulated.
+An `@info` message also prints some information, for example
+if the table has been truncated (upon request).
+
+```julia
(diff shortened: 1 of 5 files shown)


charleskawczynski:
bump 🙂

charleskawczynski:
Bump! 🙂

timholy:
I think we could have this, but I'd rather it be loaded under a `@require PrettyTables = ...` block. Often the best practice is to define the function
```
"""
Lotsa docs

Using `tabulated_invalidations` requires that you first load the PrettyTables package.
"""
function tabulated_invalidations end      # this defines a function with no methods, but the docstring is available.
```
as a "stub" in the main package, and then in `__init__()`
```julia
function __init__()
    # other stuff
    @require PettyTables = sha include("tabulated.jl")
end
```
and put the `tabulated_invalidations` methods in "tabulated.jl".

charleskawczynski:
I tried following the same sort of pattern for what's done with PyPlot, but I'm getting:

```julia
julia> using PrettyTables

julia> import SnoopCompileCore

julia> invalidations = SnoopCompileCore.@snoopr begin
           using ForwardDiff
       end;

julia> using SnoopCompile

julia> report_invalidations(; job_name = "MyWork", invalidations)
ERROR: UndefVarError: report_invalidations not defined
Stacktrace:
 [1] top-level scope
```

charleskawczynski:
Ok, this now works I think as expected, here's a mwe (assuming some packages are loadable):

```julia
import SnoopCompileCore
invalidations = SnoopCompileCore.@snoopr begin
   using ForwardDiff
end;
import SnoopCompile
using PrettyTables # needed for report_invalidations to be defined
SnoopCompile.report_invalidations(;
   invalidations,
   process_filename = x -> last(split(x, ".julia/packages/"))
)
```

Which produces

```
┌ Warning: Could not attribute the following delayed invalidations:
└ @ SnoopCompile ~/Dropbox/Caltech/work/dev/clones/forks/SnoopCompile.jl/src/invalidations.jl:552
Any[MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:init,), _A} where _A<:Tuple{AbstractDict})), MethodInstance for eltype(::Type{T} where T<:Tuple{AbstractDict}), MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:maxlog, :form), _A} where _A<:Tuple{Int64, Type})), MethodInstance for eltype(::Type{T} where T<:Tuple{Int64, Type})] invalidated Core.MethodInstance[MethodInstance for StaticArrays.var"#s25#170"(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any), MethodInstance for StaticArrays._precompile_()]
Any[MethodInstance for eltype(::Type{T} where T<:Tuple{AbstractDict}), MethodInstance for eltype(::Type{A} where A<:(NamedTuple{(:init,), _A} where _A<:Tuple{AbstractDict}))] invalidated MethodInstance for MacroTools.allbindings(::Any, ::Vector{Any})
[ Info: 233 methods invalidated for 6 functions
┌─────────────────────────────────────────────┬───────────────┬───────────────┬─────────────────┐
│ <file name>:<line number>                   │ Function Name │ Invalidations │ Invalidations % │
│                                             │               │    Number     │     (xᵢ/∑x)     │
├─────────────────────────────────────────────┼───────────────┼───────────────┼─────────────────┤
│ ForwardDiff/QdStj/src/dual.jl:427           │ promote_rule  │      209      │       82        │
│ StaticArrays/x7lS0/src/abstractarray.jl:150 │    similar    │      18       │        7        │
│ StaticArrays/x7lS0/src/SizedArray.jl:88     │    convert    │      15       │        6        │
│ StaticArrays/x7lS0/src/mapreduce.jl:253     │   foldl##kw   │       8       │        3        │
│ ForwardDiff/QdStj/src/dual.jl:145           │    isequal    │       3       │        1        │
│ StaticArrays/x7lS0/src/SOneTo.jl:57         │  getproperty  │       3       │        1        │
└─────────────────────────────────────────────┴───────────────┴───────────────┴─────────────────┘
```

charleskawczynski:
Based on https://github.com/JuliaDiff/ForwardDiff.jl/issues/620#issuecomment-1370119947 and https://github.com/JuliaDiff/ForwardDiff.jl/issues/620#issuecomment-1370142934, should we update how `report_invalidations` works? cc @ChrisRackauckas @timholy

src/invalidations.jl
```diff
@@ -255,6 +255,41 @@ else
...
+ - `job_name::String` the name of the job
+
+An `@info` message also prints some information, for example
+if the table has been truncated (upon request).
+
+```julia
+import SnoopCompileCore
+invalidations = SnoopCompileCore.@snoopr begin
+
+    # load packages & do representative work
```
timholy, line 277:
```suggestion
    # load packages & define any additional methods
```
"Doing work" cannot invalidate; invalidations are caused by new method definitions, which includes loading packages.

test/snoopr.jl
```diff
@@ -187,6 +188,14 @@ end
...
         end
     end
+
+    # Tabulate invalidations:
+    # Note: this table may change over time as invalidations get fixed.
+    SnoopCompile.report_invalidations(;
+        job_name = "job_name",
+        invalidations = invs,
+    )
+
```
timholy, line 198: This only tests that it doesn't crash. Can we print to `io = IOBuffer()` and then test the output? It could be quite loose, e.g., `occursin("Invalidations %", str)` where `str = String(take!(io))`.

src/invalidations.jl
```diff
@@ -255,6 +255,41 @@ else
...
+    report_invalidations(;
+        job_name::String = "",
+        invalidations,
+        n_rows::Int = 10,
+        process_filename::Function = x -> x,
+    )
+
+Print a tabular summary of invalidations given:
+
+ - `job_name::String` the name of the job
```
timholy, line 268: What's the role of the job_name? Does it have any functional significance? (Can I choose a name that's wrong or broken?)

src/invalidations.jl
```diff
@@ -255,6 +255,41 @@ else
...
+
+An `@info` message also prints some information, for example
+if the table has been truncated (upon request).
```
timholy, line 271: This might be interpreted as saying "you have to request the `@info` message" whereas I think you mean "it was truncated by request," which in turn really means "it was truncated by specifying `nrows`."

But even this is confusing since `nrows` has a default value of 10, so you don't actually have to request it.

src/invalidations.jl
```diff
@@ -255,6 +255,41 @@ else
...
+
+```julia
```
timholy, line 273: You might want to preface this with `Demo:` or `Usage example:`

timholy (commented):
This looks nearly merge-worthy. I don't think there are big changes needed in how you collect the invalidations, because:
1. You suggest collecting them with SnoopCompileCore rather than SnoopCompile ✔️ 
2. Nothing here uses a workload. Of course this summary will include many invalidations that may not affect the users' code. `filtermod` and `precompile_blockers` can restrict them to ones that affect particular modules or workloads. Are you wondering if you should support them?

src/invalidations.jl
charleskawczynski: Good point. I think I phrased it that way to be safe. I think I could use some clarification. If someone's "workload" was, for example:

```julia
function do_work(A)
    foo(a::Array) = a .+ 1
    x = zero(eltype(A))
    for i in 1:2
        x += sum(foo(A))
    end
    return x
end
```
Can `foo` in this example possibly invalidate methods? If so, then I think users may find `do representative work` to be a safety net to catch unexpected invalidations in real world applications.

Though, it's probably important to note that the "amount of work" (e.g., the size of `A` in this example) has no impact on the invalidations.

Overall, I think `define any additional methods` is more precise, it just requires that users think carefully about where / when they may be defining methods, which I think is fair. I'll commit this change, but I'd appreciate if you can chime in regarding my example. Thanks!

src/invalidations.jl
charleskawczynski: You know, I think we can get rid of it. It's not really doing anything. I added it in case I'd run several invalidation jobs, so that the tables could be differentiated, but 1) that can be printed before the call to `report_invalidations` and 2) I think they'd need to be separate sessions anyway. I'll remove it

test/snoopr.jl
charleskawczynski: Ah, yes, that'd be better.

charleskawczynski:
I think one last thing that we could do, that would really put this over the top in terms of ease of use, would be if we followed through with https://github.com/CliMA/ReportMetrics.jl/issues/7#issuecomment-1015618744, and made the file / line number hypertext links through PrettyTables (which I think is supported [here](https://ronisbr.github.io/PrettyTables.jl/dev/lib/library/#PrettyTables.UrlTextCell)).

I think this can be a followup feature, though. Thumbs up to open an issue 🙂

src/invalidations.jl
timholy: > If someone's "workload" was

That's not a workload, that's a method definition. `do_work(rand(5))` is a workload. Perhaps we need to clarify that workload means *running* code.

> Can foo in this example possibly invalidate methods?

Inner methods can't invalidate. But `do_work` could invalidate, if there are other methods.

timholy:
Thanks! This is a nice addition. Release of SnoopCompile 2.10.0 is underway.

PR #309, opened 2022-10-20
WIP - Reentrant concurrent snoopi_deep profiles.

NHDaly:
*NOTE:* It turns out this PR isn't mergeable as-is, since it depends on assumptions in Julia that aren't valid.
Namely, we need https://github.com/JuliaLang/julia/pull/47258 and https://github.com/JuliaLang/julia/pull/47260 to avoid data races, but it turns out those don't make sense against julia master/1.9, because of recent changes to the inference lock.

At RAI, we have backported those two PRs against 1.8.2, and we're going to keep hacking on this branch to pull it into our internal build. We will follow-up with actual changes to julia 1.9 and SnoopCompile to try to make this work for real.

--------

Following option 3 from this doc:
https://docs.google.com/document/d/1JO43w3V1Cwtb2CxWf4sYfAB68xbkTrCMr973iaJOFwY/edit#

This is a WIP PR to support invoking `@snoopi_deep` concurrently and in parallel from multiple Tasks.
(It currently requires https://github.com/JuliaLang/julia/pull/47258 to avoid data races, and https://github.com/JuliaLang/julia/pull/47260 to allow reentrant inference timing.)

The idea is basically to share the global julia type inference profile buffer, rather than clear it at the start of every profile, keeping track of the start and stop index for each invocation of snoopi_deep. Then we also clear the beginning of the profile buffer whenever the oldest invocation ends, to ensure we aren't leaking memory.


NHDaly:
Note that of course one must be careful when interpreting the results of concurrently running profiles, since inference caused by invocation B will also show up in the profile for invocation A. But this is true of all of julia's profiling and timing techniques right now, so it's just something the developer must be aware of. (Compare it with `@time` from multiple tasks reporting allocations caused by other tasks, or reporting compilation time from other tasks.)

NHDaly:
Update: it turns out that https://github.com/JuliaLang/julia/pull/47258 and https://github.com/JuliaLang/julia/pull/47260 are no longer valid, now that we're shrinking the inference lock (which is 🎉 in general though of course!).

So this PR will have to wait, because we'll probably need to rethink the inference profiler entirely, given that it currently assumes single-threaded, locked access to the inference profile buffer. :/

NHDaly:
CC: @sacha0 for review plz

test/snoopi_deep.jl
```diff
@@ -863,6 +863,7 @@ end
     # pgdsgui(axs[2], rit; bystr="Inclusive", consts=true, interactive=false)
 end
 
+
```
Sacha0, line 866: Snuck in? :)
```suggestion
```

test/snoopi_deep.jl
```diff
@@ -944,6 +945,163 @@ end
     Pkg.activate(cproj)
 end
 
+_name(frame::SnoopCompileCore.InferenceTiming) = frame.mi_info.mi.def.name
+
+@testset "reentrant concurrent profiles 1 - overlap" begin
+    # Warmup
```
Sacha0, line 951: Perhaps something like the following? :)
```suggestion
    # Warmup to prevent, e.g., `+` from appearing in the profiles; only `ROOT` and `foo*` should appear.
```

Sacha0: (Likewise on the warmup comments below if you agree.)

test/snoopi_deep.jl
```diff
@@ -944,6 +945,163 @@ end
...
+    @test isempty(SnoopCompileCore.SnoopiDeepParallelism.invocations)
+    @test isempty(Core.Compiler.Timings._timings[1].children)
+end
+
+@testset "reentrant concurrent profiles 3 - nested" begin
+    # Warmup
+    @eval foo1(x) = x+2
+    @eval foo1(2)
+
+    # Test:
```
Sacha0, line 1026: That this works is thoroughly pleasing.

test/snoopi_deep.jl
```diff
@@ -944,6 +945,163 @@ end
...
+    snoop_times = Float64[0.0, 0.0, 0.0, 0.0]
+    # Run it twice to ensure we warmup the eval block
+    for _ in 1:2
+        @sync begin
+            ts = [
+                Threads.@spawn begin
+                    sleep((i-1) / 10)  # (Divide by 10 so the test isn't too slow)
+                    snoop_time = @timed SnoopCompile.@snoopi_deep @eval begin
+                        $(Symbol("foo$i"))(x) = x + 1
+                        sleep(1.5 / 10)
```
Sacha0, line 1070: I hope that these delays are large enough --- and think they should be by orders of magnitude --- to prevent nondeterministic failures due to jitter. But I've found the machines are just lying in wait, for the moment I think something like that. Then 🦈 🩸.

test/snoopi_deep.jl
```diff
@@ -944,6 +945,163 @@ end
...
+    # Test the sanity of the reported Timings
+    @testset for i in eachindex(profs)
+        prof = profs[i]
+        # Test that the time for the inference is accounted for
+        @test 0.15 < prof.mi_timing.exclusive_time
+        @test prof.mi_timing.exclusive_time < prof.mi_timing.inclusive_time
+        # Test that the inclusive time (the total time reported by snoopi_deep) matches
+        # the actual time to do the snoopi_deep, as measured by `@time`.
+        # These should both be approximately ~0.15 seconds.
+        @info prof.mi_timing.inclusive_time
```
Sacha0, line 1096: Debugging straggler? :)

test/snoopi_deep.jl
```diff
@@ -944,6 +945,163 @@ end
...
+    @test Set(_name.(SnoopCompile.flatten(profs[1]))) == Set([:ROOT, :foo1])
+    @test Set(_name.(SnoopCompile.flatten(profs[2]))) == Set([:ROOT, :foo1, :foo2])
+    @test Set(_name.(SnoopCompile.flatten(profs[3]))) == Set([:ROOT,        :foo2, :foo3])
+    @test Set(_name.(SnoopCompile.flatten(profs[4]))) == Set([:ROOT,               :foo3, :foo4])
+
+    # Test the sanity of the reported Timings
+    @testset for i in eachindex(profs)
+        prof = profs[i]
+        # Test that the time for the inference is accounted for
+        @test 0.15 < prof.mi_timing.exclusive_time
```
Sacha0, line 1091: I fear the definitions of these times has already become hazy to me 😅. This one? 🤔How do know that the exclusive rather than inclusive time should be over 0.15 seconds?

SnoopCompileCore/src/snoopi_deep.jl
```diff
@@ -70,28 +70,201 @@ function addchildren!(parent::InferenceTimingNode, t::Core.Compiler.Timings.Timi
...
+function _current_profile_stats_locked()
+    ccall(:jl_typeinf_lock_begin, Cvoid, ())
+    try
+        inference_root_timing = Core.Compiler.Timings._timings[1]
+        children = inference_root_timing.children
+        # Since we were able to grab the lock, we must not be in an inference profile,
+        # meaning we are in ROOT(). So to get an accurate ROOT timing, we have to add the
+        # accumulated time since the ROOT was last updated:
+        accum_root_time = time_ns() - inference_root_timing.cur_start_time
+        current_root_time = inference_root_timing.time + accum_root_time
```
Sacha0, line 128: I'm a bit hazy on these details by now but trust that it is correct 👍

NHDaly:
Thanks again for the review, Sacha. I'm closing this for now, since we ended up going a different approach, via https://github.com/JuliaLang/julia/pull/47615 and https://github.com/timholy/SnoopCompile.jl/pull/313.

PR #310, opened 2022-10-28
Move `flatten()` to SnoopCompileCore.

NHDaly:
At RelationalAI, we are turning on always-on `@snoopi_deep` profiling in our production servers. For this, we want to add a dependency on SnoopCompileCore, but not on SnoopCompile, since SnoopCompile adds many dependencies for visualization utilities, such as FlameGraphs.jl.

We think that `flatten()` is a small enough and simple enough utility function that it could easily live in SnoopCompileCore. Does that sound reasonable to you as well?

Thanks! :)


NHDaly:
CC: @sacha0 and @vilterp for review.

NHDaly:
🤔 the test failures look unrelated, from what i can see:

> 
> ```julia
> snoopi: Test Failed at /Users/runner/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:132
>   Expression: any((str->begin
>             occursin("kwftype", str)
>         end), FK)
> Stacktrace:
>  [1] macro expansion
>    @ ~/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.9/Test/src/Test.jl:477 [inlined]
>  [2] macro expansion
>    @ ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:132 [inlined]
>  [3] macro expansion
>    @ ~/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.9/Test/src/Test.jl:1496 [inlined]
>  [4] top-level scope
>    @ ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:97
> ┌ Warning: Body method was not toplevel-inferred, test skipped
> └ @ Main ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:143
> snoopi: Test Failed at /Users/runner/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:153
>   Expression: any((str->begin
>             occursin("kwftype", str)
>         end), FK)
> Stacktrace:
>  [1] macro expansion
>    @ ~/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.9/Test/src/Test.jl:477 [inlined]
>  [2] macro expansion
>    @ ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:153 [inlined]
>  [3] macro expansion
>    @ ~/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.9/Test/src/Test.jl:1496 [inlined]
>  [4] top-level scope
>    @ ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:97
> snoopi: Test Failed at /Users/runner/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:[182](https://github.com/timholy/SnoopCompile.jl/actions/runs/3347452311/jobs/5545449984#step:8:185)
>   Expression: any((str->begin
>             occursin("Tuple{Core.kwftype(typeof(genkw2)),NamedTuple{(:b,),$(SP)Tuple{String}},typeof(genkw2)}", str)
>         end), FK)
> Stacktrace:
>  [1] macro expansion
>    @ ~/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.9/Test/src/Test.jl:477 [inlined]
>  [2] macro expansion
>    @ ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:182 [inlined]
>  [3] macro expansion
>    @ ~/hostedtoolcache/julia/nightly/x64/share/julia/stdlib/v1.9/Test/src/Test.jl:1496 [inlined]
>  [4] top-level scope
>    @ ~/work/SnoopCompile.jl/SnoopCompile.jl/test/snoopi.jl:97
> ```

vilterp (approved):

timholy:
Not quite sure what to do here. I was trying to keep SnoopCompileCore to be about data-collection only.

But I get why you'd want something slimmer. The key principle is that SnoopCompileCore can't extend any Base methods (it cannot itself be a source of any invalidations), and obviously this doesn't violate that. My only concern is that this opens the door to moving other things, too, and the end result could be confusing to users. For example, the printing from `flatten` is probably pretty awful without the `show` methods, right? But the `show` methods can't move to SnoopCompileCore.

I'll think about this for another day or two and then decide about merging.

NHDaly:
Yeah, Tim, I had similar concerns.. We ended up just extracting this method out into our own codebase, which seems fine for now. I think it's up to you about how to proceed. 🤷

timholy:
I think in the absence of strong pressure to move it, I'd rather keep the clean "collection vs analysis" separation. But I'm happy to continue to brainstorm about this as we move forward. I totally get where you're coming from on this.

NHDaly:
sounds right, thanks! :)

PR #311, opened 2022-11-17
[WIP] use new timing api

vilterp:
introduced in https://github.com/JuliaLang/julia/pull/47615

cc @NHDaly


NHDaly:
I think this can be closed in favor of https://github.com/timholy/SnoopCompile.jl/pull/313 now, yeah?

PR #313, opened 2022-11-22
Adapt SnoopPrecompile & snoopi_deep to work for new julia API.

vilterp:
Starting in https://github.com/JuliaLang/julia/pull/47615, we are changing the API for snoopi_deep, to allow for thread safe snooping in one thread while compiling in another.

This commit changes SnoopPrecompile to work with the new API if it's available.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "SnoopCompile"
 uuid = "aa65fe97-06da-5843-b5b1-d5d13cad87d2"
 author = ["Tim Holy <tim.holy@gmail.com>"]
-version = "2.9.5"
+version = "2.9.6"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/SnoopPrecompile/Project.toml
+++ b/SnoopPrecompile/Project.toml
@@ -1,7 +1,7 @@
 name = "SnoopPrecompile"
 uuid = "66db9d55-30c0-4569-8b51-7e840670fc0c"
 authors = ["Tim Holy <tim.holy@gmail.com> and contributors"]
-version = "1.0.1"
+version = "1.0.2"
 
 [compat]
 julia = "1"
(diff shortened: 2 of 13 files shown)


SnoopCompileCore/src/snoopi_deep.jl
```diff
@@ -70,17 +74,34 @@ function addchildren!(parent::InferenceTimingNode, t::Core.Compiler.Timings.Timi
...
+        Core.Compiler.__set_measure_typeinf(false)
+    end
+    function finish_snoopi_deep()
+        # Construct a dummy Timing for ROOT(), for backwards compatibility with the old API.
+        root = Timing(
+            InferenceFrameInfo(ROOTmi, 0x0, Any[], Any[Core.Const(ROOT)], 1),
+            0x0)
+        root.children = Core.Compiler.Timings.clear_and_fetch_timings()
+
+        return InferenceTimingNode(root)
```
NHDaly, line 91: @timholy: In https://github.com/JuliaLang/julia/pull/47615, we changed the API to return a _vector of inference trees_, rather than a single inference tree with the dummy "ROOT" node.

We think this is a more sensical API, since there really _wasn't_ a single tree of type inference, there were many, and it also made it easier to get the thread safety correct.

However, of course this breaks the old API. So for now, I'm proposing we just do this, and shove it back into a dummy ROOT node to make a single tree 😅 

Then we can investigate changes to the API in the future if we want to clean things up to all operate on vectors of trees, or just leave it like this forever. :)

Does that sound good to you?

NHDaly (approved):
I think this LGTM. 👍 The tests pass on our local branch of https://github.com/JuliaLang/julia/pull/47615, and on a release build, so I think this is good to go.

Thanks @vilterp. 👍

NHDaly:
There are some other test failures here, but they seem very much unrelated. If anyone (@timholy?) can comment on it, that'd be appreciated, otherwise I plan to merge this PR next week. :)

timholy:
This will need to be updated for SnoopCompile 3, if it's still relevant. See #382 and especially #384.

Issue #317, opened 2022-12-15
latest version of `SnoopPrecompile` breaks `RecipesBase`


t-bltg:
@timholy, https://github.com/JuliaPlots/Plots.jl/issues/4597.

timholy:
By way of explanation, see https://github.com/JuliaLang/julia/issues/47902

t-bltg:
No harm done here, stuff like that happen ;)
I have a simplified reproducer, will open a PR in a moment.

timholy:
Did you test whether putting it in the `@precompile_setup` block worsened your outcomes? I'm loathe to add restrictions but we could in SnoopPrecompile 2, if it improved outcomes elsewhere.

There's no immediate rush on this, and indeed I'm comfortable with waiting a bit. The bigger picture is that I'm currently tracking down cases in which the native-code caching doesn't give as much of a benefit as I think it should. SnoopPrecompile 1.0.2 was supposed to be part of the solution, but it's clearly not the whole solution. Since I have that locally and can use it in trying to figure out what the other causes are, it's fine if we wait a bit to see what additional clarity about root causes brings.

t-bltg:
Let's continue the discussion in https://github.com/timholy/SnoopCompile.jl/pull/318#issue-1498437547.

PR #318, opened 2022-12-15
revert broken commit - add test

t-bltg:
@timholy, the added test breaks on `1.0.2` and not on `1.0.1`.

So I've reverted the changes of `1.0.2` and we can start working on a fix.

I'm not sure if the precompilation scenario is valid code in `RecipesBase` (this is clearly a niche involving meta-programming), and I'd be happy to change some - considered invalid - but working statements in https://github.com/JuliaPlots/Plots.jl/blob/master/RecipesBase/src/RecipesBase.jl#L600-L619.

### DIFF
--- a/SnoopPrecompile/src/SnoopPrecompile.jl
+++ b/SnoopPrecompile/src/SnoopPrecompile.jl
@@ -60,11 +60,10 @@ macro precompile_all_calls(ex::Expr)
     end
     if have_inference_tracking
         ex = quote
-            thunk() = $ex
             Core.Compiler.Timings.reset_timings()
             Core.Compiler.__set_measure_typeinf(true)
             try
-                Base.invokelatest(thunk)
+                $ex
             finally
                 Core.Compiler.__set_measure_typeinf(false)
                 Core.Compiler.Timings.close_current_timer()
@@ -105,11 +104,11 @@ to your package).
 """
 macro precompile_setup(ex::Expr)
     return esc(quote
-        let
+        # let
             if ccall(:jl_generating_output, Cint, ()) == 1 || $SnoopPrecompile.verbose[]
                 $ex
             end
-        end
+        # end
     end)
 end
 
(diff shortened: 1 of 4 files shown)


timholy:
Since SnoopPrecompile 1.0 supports putting a method inside `@precompile_all_calls`, we have to support that at least until SnoopPrecompile 2.0. This test is extremely useful, thanks!

t-bltg:
Let me know if you need something else, I'm not really into the `precompilation` internals but if I can help in any way :)

timholy:
Continuing from #317, the main issue is how to ensure that some of the workload in the `@precompile_all_calls` doesn't get compiled before inference starts tracking it. xref https://github.com/JuliaLang/julia/issues/47902. That issue details potential solutions. If you're wanting to make sure that the macro-expansion itself gets fully precompiled, my sense is the `@eval` solution might be the right one. But I think that would require you to use `$__RecipesBasePrecompileType` anywhere you current use it without the `$` currently. That of course is itself a breaking change.

timholy:
Basically is your goal to ensure everything needed by `@recipe` gets precompiled? AFAICT that only helps reduce package precompilation times, it doesn't affect actual usage of the packages that use RecipesBase.

timholy:
I'm guessing there's another `let` introduction in 1.0.2 that needs to be reverted...

t-bltg:
> Basically is your goal to ensure everything needed by @recipe gets precompiled? AFAICT that only helps reduce package precompilation times, it doesn't affect actual usage of the packages that use RecipesBase.

Yeah, that probably and overkill to use `SnoopPrecompile` there ...

timholy:
I wouldn't necessarily call it overkill, I totally see why you'd do that. Just trying to come up with a solution that is as transparent to users about what will and what won't work. There's a real expectation that lines will be executed in the order in which they appear, but without special precautions that's not actually true.

t-bltg:
Hum, the pseudo `RecipeBase` precompilation code might not even be valid in julia versions `<= 1.4` ...

When I added the `SnoopPrecompile` dep to `RecipesBase`, we restricted the julia version to `>= 1.6`.

timholy:
I think it would be OK to restrict that test to Julia versions where it works. In my mind the point is to make sure we catch potential regressions; if it never worked in any SnoopPrecompile release, then we don't have to figure out how to make it work.

t-bltg:
`nightly` failure seems unrelated (`JET` failure).

timholy:
Yep, see #300

timholy:
Thanks for a very important contribution!

PR #326, opened 2022-12-30
Add a copy-paste invalidations analysis script

ChrisRackauckas:
As much as details are nice, sometimes giving something that's easy to copy/repeat is good. Maybe this could be added as a library function (it probably would need to be a macro).

### DIFF
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -113,7 +113,7 @@ with few children, and thus this shows how many "bad actors" need to be investig
 invalidations, with `show(trees[end-1])` being the second most, and so forth.
 
 ```julia
-using SnoopCompile
+using SnoopCompileCore
 invalidations = @snoopr begin
     # Your code goes here!
     using OrdinaryDiffEq
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -134,7 +134,7 @@ trees = SnoopCompile.invalidation_trees(invalidations);
 
 @show length(SnoopCompile.uinvalidated(invalidations)) # show total invalidations
 
-show(trees[end]) # show the most invalidated method
+show(trees[end]) # show the most invalidating method
 
 # Count number of children (number of invalidations per invalidated method)
 n_invalidations = map(trees) do methinvs
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -6,8 +6,11 @@ You can find a more expansive version of this page in [this blog post](https://j
 
 ## Cut to the Chase: A copy-paste analysis of invalidations
 
-The following is a quick "grab and go" script for analyzing invalidations. Insert your code into the `@snoopr` block. The resulting plot shows the
-distributions of the invalidations sorted by the number of children affected. Generally, invalidations with many children matter more than those
+The following is a quick "grab and go" script for analyzing invalidations.
+Insert package loads (`using` or `import` statements) and/or method definitions into the `@snoopr` block,
+and put the workload you want to be fast in the `@snoopi_deep` block.
+The resulting plot shows the distribution of the invalidations sorted by the number of children affected.
+Generally, invalidations with many children matter more than those
 with few children, and thus this shows how many "bad actors" need to be investigated. `show(trees[end])` show the method which leads to the most
 invalidations, with `show(trees[end-1])` being the second most, and so forth.
 
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -18,11 +18,11 @@ While the plot shows total invalidations (`trees`), only the ones in `staletrees
 ```julia
 using SnoopCompileCore
 invalidations = @snoopr using PkgA, PkgB;
-using SnoopCompile
-trees = invalidation_trees(invalidations)
 tinf = @snoopi_deep begin
     some_workload()
 end
+using SnoopCompile
+trees = invalidation_trees(invalidations)
 staletrees = precompile_blockers(trees, tinf)
 
 @show length(SnoopCompile.uinvalidated(invalidations)) # show total invalidations
--- a/docs/src/tutorial.md
+++ b/docs/src/tutorial.md
@@ -105,6 +105,52 @@ because `Vector{AbstractFloat}` is a concrete type, whereas `AbstractFloat` is n
 
 If we create `c32 = [1.0f0]` and then `calldouble2(c32)`, we would also see backedges from `double(::Float32)` all the way back to `calldouble2(::Vector{Float32})`.
 
+## A copy-paste analysis of invalidations
+
+The following is a quick "grab and go" script for analyzing invalidations. Insert your code into the `@snoopr` block. The resulting plot shows the
+distributions of the invalidations sorted by the number of children effected. Generally, invalidations with many children matter more than those
+with few children, and thus this shows how many "bad actors" need to be investigated. `show(trees[end])` show the method which leads to the most
+invalidations, with `show(trees[end-1])` being the second most, and so forth.
+
+```julia
+using SnoopCompile
+invalidations = @snoopr begin
+    # Your code goes here!
+    using OrdinaryDiffEq
+
+    function lorenz(du, u, p, t)
+        du[1] = 10.0(u[2] - u[1])
+        du[2] = u[1] * (28.0 - u[3]) - u[2]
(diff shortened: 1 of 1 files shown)


docs/src/tutorial.md
```diff
@@ -105,6 +105,52 @@ because `Vector{AbstractFloat}` is a concrete type, whereas `AbstractFloat` is n
...
 
+## A copy-paste analysis of invalidations
+
+The following is a quick "grab and go" script for analyzing invalidations. Insert your code into the `@snoopr` block. The resulting plot shows the
+distributions of the invalidations sorted by the number of children effected. Generally, invalidations with many children matter more than those
+with few children, and thus this shows how many "bad actors" need to be investigated. `show(trees[end])` show the method which leads to the most
+invalidations, with `show(trees[end-1])` being the second most, and so forth.
+
+```julia
+using SnoopCompile
```
timholy, line 116:
```suggestion
using SnoopCompileCore
```
SnoopCompile triggers quite a few invalidations of its own and can hide problems.

docs/src/tutorial.md
```diff
@@ -105,6 +105,52 @@ because `Vector{AbstractFloat}` is a concrete type, whereas `AbstractFloat` is n
...
+invalidations = @snoopr begin
+    # Your code goes here!
+    using OrdinaryDiffEq
```
timholy, line 119: Unless there's already a `lorenz` function lying around, `using OrdinaryDiffEq` is the only thing here that can cause invalidations. The rest is a bit misleading because it implies that inference/compilation itself can cause invalidations. (It can't.)

docs/src/tutorial.md
```diff
@@ -105,6 +105,52 @@ because `Vector{AbstractFloat}` is a concrete type, whereas `AbstractFloat` is n
...
+    prob = ODEProblem{true,false}(lorenz, u0, tspan)
+    alg = Rodas5()
+    tinf = solve(prob, alg)
+end;
+
+trees = SnoopCompile.invalidation_trees(invalidations);
+
+@show length(SnoopCompile.uinvalidated(invalidations)) # show total invalidations
+
+show(trees[end]) # show the most invalidated method
```
timholy, line 137:
```suggestion
show(trees[end]) # show the most invalidating method
```

timholy (commented):
This looks like a good idea. I made some specific recommendations. However, I think the recipe in https://discourse.julialang.org/t/julia-v1-9-0-beta2-is-fast/92290/17?u=tim.holy has several merits, most particularly that by using `precompile_blockers` you focus just on the invalidations that affect your workload. (*I* tend to fix invalidations wherever I find them, but I think for many developers they don't want to be worried about, say, REPL invalidations when their goal is to reduce TTFX for their package.)

To turn it into a more fleshed-out example of my schematic recipe, you could put the `using OrdinaryDiffEq` and method definition under `@snoopr` (those are the parts that can cause invalidation) and the rest (the parts that may trigger inference) under `@snoopi_deep` and then use `precompile_blockers`.

docs/src/tutorial.md
ChrisRackauckas: There are weird eval cases where it can, like if you change a Plots backend and it evals in some new code. But that might be too rare to specialize on.

docs/src/tutorial.md
```diff
@@ -105,6 +105,52 @@ because `Vector{AbstractFloat}` is a concrete type, whereas `AbstractFloat` is n
...
+        du[1] = 10.0(u[2] - u[1])
+        du[2] = u[1] * (28.0 - u[3]) - u[2]
+        du[3] = u[1] * u[2] - (8 / 3) * u[3]
+    end
+    u0 = [1.0; 0.0; 0.0]
+    tspan = (0.0, 100.0)
+    prob = ODEProblem{true,false}(lorenz, u0, tspan)
+    alg = Rodas5()
+    tinf = solve(prob, alg)
+end;
```
ChrisRackauckas, line 131:
```suggestion
invalidations = @snoopr begin
    # Your packages go here!
    using OrdinaryDiffEq
end;
```

docs/src/tutorial.md
```diff
@@ -105,6 +105,41 @@ because `Vector{AbstractFloat}` is a concrete type, whereas `AbstractFloat` is n
 
 If we create `c32 = [1.0f0]` and then `calldouble2(c32)`, we would also see backedges from `double(::Float32)` all the way back to `calldouble2(::Vector{Float32})`.
 
+## A copy-paste analysis of invalidations
+
+The following is a quick "grab and go" script for analyzing invalidations. Insert your code into the `@snoopr` block. The resulting plot shows the
+distributions of the invalidations sorted by the number of children effected. Generally, invalidations with many children matter more than those
```
ChrisRackauckas, line 111:
```suggestion
distributions of the invalidations sorted by the number of children affected. Generally, invalidations with many children matter more than those
```

ChrisRackauckas:
I'll go by your recommendations, but I think my main points are:

1. I think the tutorial should have a simple copy paste "what are my invalidations?" before the deeper explanation of the tool
2. It should also produce a plot to show the relative scale of which invalidations matter.

So I updated to make use of the version from https://timholy.github.io/SnoopCompile.jl/dev/reference/#SnoopCompile.precompile_blockers. For some reason I find that way it's written to be much easier to read than the one in https://discourse.julialang.org/t/julia-v1-9-0-beta2-is-fast/92290/17?u=tim.holy, maybe just due to white space. 

For the plots, should I be doing that all on `staletrees` instead of `trees`?

ChrisRackauckas:
Maybe this should be in a separate "Getting Started" or "SnoopCompiler 101 for Inexperienced Programmers" or something. I think the big thing about SnoopCompile is that now with the precompilation changes and all, it needs to move from an audience of "I know a bit of how Julia works and want to make a good package better" to "this is some standard Julia 201 stuff like type-stability". I don't think your average "just started developing packages" developer should need to know how to dig into the details and fix invalidations. I think it's sufficient for many to just know that (a) invalidations are an issue, (b) here's how I create a report of invalidations and open an issue. If people are opening these issues and the big ones are brought into the conversation, then I think they will be solved.

timholy:
I agree this has to get a lot simpler. Most of the docs were written in the course of this basically being a research project, trying to understand what factors would affect precompilation. Now that Julia itself is much more capable, this is a much less complex question than it used to be. We definitely need to make the documentation reflect the new world.

In fact what I think we also desperately need is a way to take the analysis that Cthulhu does, but instead of annotating the type-inferred code, instead annotates the raw source text. E.g.,
```
    for (i, x::Any) in enumerate(v)
        g[i] = f(x)::Any
    end
```
but making that happen is, I think, quite difficult. https://github.com/JuliaLang/julia/issues/31162 would make this vastly easier, but I don't know lisp and have no intention of learning.

docs/src/tutorial.md
```diff
@@ -4,6 +4,42 @@ Certain concepts and types will appear repeatedly, so it's worth
...
+    some_workload()
+end
+staletrees = precompile_blockers(trees, tinf)
+
+@show length(SnoopCompile.uinvalidated(invalidations)) # show total invalidations
+
+show(trees[end]) # show the most invalidating method
+
+# Count number of children (number of invalidations per invalidated method)
+n_invalidations = map(trees) do methinvs
```
timholy, line 29: So I guess the question is, do we want to plot *all* invalidations or just the ones that affect the workload? I can go with this, I just worry it might seem daunting.

timholy (commented):
Just a couple more things to think about and then I think we can merge this.

docs/src/tutorial.md
```diff
@@ -4,6 +4,42 @@ Certain concepts and types will appear repeatedly, so it's worth
...
+    SnoopCompile.countchildren(methinvs)
+end
```
ranocha, line 31: Isn't this just
```suggestion
n_invalidations = map(SnoopCompile.countchildren, trees)
```
(without forming an unnecessary anonymous function)?

docs/src/tutorial.md
ranocha: What about plotting both next to each other?

timholy: Or in different colors on the same axis.

timholy:
I think we can go with this, and add other forms of plotting later. Thanks @ChrisRackauckas! I'll merge when CI finishes.

PR #330, opened 2023-01-03
Fix tests for introduction of `Core.kwcall`

timholy:
This also resolves some issues with looking up bodyfunction names.


timholy:
The failure in `Stale` will be fixed by #329.

Issue #324, opened 2022-12-28
disable precompilation on `JULIA_PKG_PRECOMPILE_AUTO` ?


t-bltg:
Just a thought here, should we disable `SnoopPrecompile` precompilation on `JULIA_PKG_PRECOMPILE_AUTO` env variable ?

I'd like to disable precompilation when developing e.g. `Makie`.

timholy:
I think Preferences.jl would be a better approach, see #299. We just got the ability to trigger PkgEval runs from packages, which should let us avoid the fiasco from the last time I tried to enhance SnoopPrecompile 🙂. I'm also planning on transferring SnoopPrecompile to JuliaLang, since it's pretty core infrastructure.

t-bltg:
> I think Preferences.jl would be a better approach, see https://github.com/timholy/SnoopCompile.jl/pull/299

I see, thanks for the pointers. See the naive implementation at https://github.com/timholy/SnoopCompile.jl/pull/334.

> I'm also planning on transferring SnoopPrecompile to JuliaLang, since it's pretty core infrastructure.

Great !

PR #334, opened 2023-01-08
disable precompilation via `Preferences`

t-bltg:
Fix https://github.com/timholy/SnoopCompile.jl/issues/324.


t-bltg:
Hum, `ERROR: LoadError: Preferences.jl can only be used on Julia v1.6+!`.

timholy:
I wonder if we should make it more complicated 🙂 . Specifically, would it be useful to allow one to specify a list of packages one shouldn't execute the precompile workload for? If I'm developing a specific package I still might want good precompilation on others.

I'm happy to help with this. I was going to transfer SnoopPrecompile first but with this as a head-start there's probably no reason to do so.

t-bltg:
> Specifically, would it be useful to allow one to specify a list of packages one shouldn't execute the precompile workload for?

Done.

docs/src/snoop_pc.md
```diff
@@ -80,3 +80,11 @@ intended calls from inside the `@precompile_all_calls` block (see [`@snoopi_deep
     code that you want precompiled.
     You can use `@snoopi_deep` to check for any (re)inference when you use the code in your package.
     To fix any specific problems, you can combine `@precompile_all_calls` with manual `precompile` directives.
+
+One can disable precompilation for selected packages using the `Preferences.jl` based mechanism and the `skip_precompile` key:
```
timholy, line 84:
```suggestion
One can reduce the cost of precompilation for selected packages using the `Preferences.jl` based mechanism and the `skip_precompile` key:
```

docs/src/snoop_pc.md
```diff
@@ -80,3 +80,11 @@ intended calls from inside the `@precompile_all_calls` block (see [`@snoopi_deep
...
+```julia
+using SnoopPrecompile, Preferences
+set_preferences!(SnoopPrecompile, "skip_precompile" => ["PackageA", "PackageB"])
+```
+
+After restarting julia, precompilation will be disabled for `PackageA` and `PackageB`.
```
timholy, line 90:
```suggestion
After restarting julia, the `@precompile_all_calls` and `@precompile_setup` workloads will be disabled for `PackageA` and `PackageB`.
```

timholy (commented):
I tweaked the docs since precompilation itself isn't disabled, we're just skipping the workload that results in "extra" precompilation.

As far as a test, one could perhaps use
```
@precompile_all_calls begin
    global workload_ran = true
end
```
and check `isdefined(SomeModule, :workload_ran) && SomeModule.workload_ran`.

t-bltg:
@timholy, can you cancel the previous Documenter workflows ? (Or maybe give me contribution rights to this repo ?)

They fill up the actions queue and prevent running CI for most recent commits.
I've added a concurrency block in the `.yml` files.

docs/src/snoop_pc.md
```diff
@@ -80,3 +80,11 @@ intended calls from inside the `@precompile_all_calls` block (see [`@snoopi_deep
...
     You can use `@snoopi_deep` to check for any (re)inference when you use the code in your package.
     To fix any specific problems, you can combine `@precompile_all_calls` with manual `precompile` directives.
+
+One can reduce the cost of precompilation for selected packages using the `Preferences.jl` based mechanism and the `skip_precompile` key:
+```julia
+using SnoopPrecompile, Preferences
+set_preferences!(SnoopPrecompile, "skip_precompile" => ["PackageA", "PackageB"])
+```
+
+After restarting julia, the `@precompile_all_calls` and `@precompile_setup` workloads will be disabled (locally) for `PackageA` and `PackageB`.
```
timholy, line 90:
```suggestion
After restarting julia, the `@precompile_all_calls` and `@precompile_setup` workloads will be disabled (locally) for `PackageA` and `PackageB`.

!!! note
    Changing `skip_precompile` may result in a one-time recompilation of all packages that use SnoopPrecompile.
```

timholy (approved):

timholy:
Thanks so much!

Ideally I want to try a PkgEval run before we release this, but we should get this out soon: this meets a real need! ❤️

PR #341, opened 2023-01-27
Use active project for `snoopl`

chriselrod:
It's common to use projects and load libraries not in the base environment, or possibly ones with different versions.


timholy:
Thanks!

PR #343, opened 2023-02-09
make SnoopCompile compatible with JET@0.7

aviatesk:


aviatesk:
Ops, we already have #342.

PR #363, opened 2023-04-26
Add deprecation notice to docs on SnoopPrecompile

timholy:


timholy:
xref https://github.com/aviatesk/JET.jl/pull/498 and https://github.com/JuliaLang/PrecompileTools.jl/pull/13

PR #376, opened 2024-02-26
Turn Cthulhu and JET into weakdeps

mkitti:
Create CthulhuExt and JETExt extension packages

Cthulhu.jl and JET.jl are rather large and complicated dependencies.

By making them weak dependencies, we increase the likelihood that SnoopCompile.jl could be loaded on development branches of Julia.

At the time of this writing, SnoopCompile.jl cannot be loaded on the Julia 1.11 branch due to an issue with Cthulhu precompilation.

### DIFF
--- a/src/parcel_snoopi_deep.jl
+++ b/src/parcel_snoopi_deep.jl
@@ -1843,9 +1843,6 @@ function max_end_time(node::InferenceTimingNode, recursive::Bool=false, tmax=-on
     return tmax
 end
 
-# See ext/JETExt.jl for implementation
-function report_callees end
-
 for IO in (IOContext{Base.TTY}, IOContext{IOBuffer}, IOBuffer)
     for T = (InferenceTimingNode, InferenceTrigger, Precompiles, MethodLoc, MethodTriggers, Location, LocationTriggers)
         @warnpcfail precompile(show, (IO, T))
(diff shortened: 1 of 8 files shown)


src/parcel_snoopi_deep.jl
```diff
@@ -1912,6 +1843,9 @@ function max_end_time(node::InferenceTimingNode, recursive::Bool=false, tmax=-on
     return tmax
 end
 
+# See ext/JETExt.jl for implementation
+function report_callees end
+
```
aviatesk, line 1848:
```suggestion
```

test/runtests.jl
```diff
@@ -1,5 +1,10 @@
 using Test
 
+if Base.VERSION >= v"1.7"
```
aviatesk, line 3:
```suggestion
if VERSION >= v"1.7"
```

aviatesk:
Thanks for your PR, @mkitti ! Unfortunately, this seems to be broken on Julia versions earlier than 1.8 (pre pkg extensions), so I'd like to go with #379 first. I'd appreciate if we can merge the change to make Cthulhu integration lazy from this PR though. Do you want to update this PR, or want me to work on that?

mkitti:
Closing to rebase and refactor

PR #377, opened 2024-04-02
Update the docs to mention that native code is now cached

JamesWrigley:
I noticed that they still say that native code isn't cached.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -28,23 +28,24 @@ code generation performs optimizations and ultimately generates the assembly
 language (native code) used on CPUs.
 Some aspects of this process are documented [here](https://docs.julialang.org/en/v1/devdocs/eval/).
 
-Every time you load a package in a fresh Julia session, the methods you use
-need to be JIT-compiled, and this contributes to the latency of using the package.
-In some circumstances, you can partially cache the results of compilation to a file
-(the `*.ji` files that live in your `~/.julia/compiled` directory) to reduce the burden
-when your package is used.
-This is called *precompilation*.
-Unfortunately, precompilation is not as comprehensive as one might hope.
-Currently, Julia is only able to save inference results (not native code) in the
-`*.ji` files, and thus precompilation only eliminates the time needed for type inference.
-Moreover, there are some significant constraints that sometimes prevent Julia from
-saving even the inference results--for example, currently you cannot cache inference results
-for "top level" calls to methods defined in Julia or other packages, even if you are calling them
-with types defined in your package.
-Finally, what does get saved can sometimes be invalidated by loading other packages.
+Every time you load a package in a fresh Julia session, the methods you use need
+to be JIT-compiled, and this contributes to the latency of using the package.
+In some circumstances, you can cache the results of compilation to files to
+reduce the latency when your package is used. These files are the the `*.ji` and
+`*.so` files that live in the `compiled` directory of the Julia depot, usually
+located at `~/.julia/compiled`. However, if these files become large, loading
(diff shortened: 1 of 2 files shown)


mkitti (approved):

docs/src/index.md
```diff
@@ -30,18 +30,20 @@ Some aspects of this process are documented [here](https://docs.julialang.org/en
...
+`*.ji` files and ([since Julia
+1.9](https://julialang.org/blog/2023/04/julia-1.9-highlights/#caching_of_native_code))
+native code in the `*.so` files, and thus precompilation can eliminate the time
+needed for type inference and native code compilation.
+However, there are some significant constraints that sometimes prevent Julia from
 saving even the inference results--for example, currently you cannot cache inference results
 for "top level" calls to methods defined in Julia or other packages, even if you are calling them
 with types defined in your package.
-Finally, what does get saved can sometimes be invalidated by loading other packages.
+On top of this, what does get saved can sometimes be invalidated by loading other packages.
```
timholy, line 46: The entire sentence "However, there are some..." can also be deleted.

timholy (commented):
Thanks!

docs/src/index.md
```diff
@@ -30,18 +30,20 @@ Some aspects of this process are documented [here](https://docs.julialang.org/en
...
-Moreover, there are some significant constraints that sometimes prevent Julia from
+Julia is able to save inference results in the
+`*.ji` files and ([since Julia
+1.9](https://julialang.org/blog/2023/04/julia-1.9-highlights/#caching_of_native_code))
+native code in the `*.so` files, and thus precompilation can eliminate the time
+needed for type inference and native code compilation.
+However, there are some significant constraints that sometimes prevent Julia from
 saving even the inference results--for example, currently you cannot cache inference results
 for "top level" calls to methods defined in Julia or other packages, even if you are calling them
 with types defined in your package.
```
mkitti, line 45:
```suggestion
```

JamesWrigley:
Thanks both, I deleted that sentence and reworded some others slightly in 28ec9a3.

docs/src/index.md
```diff
@@ -30,21 +30,17 @@ Some aspects of this process are documented [here](https://docs.julialang.org/en
 
 Every time you load a package in a fresh Julia session, the methods you use
 need to be JIT-compiled, and this contributes to the latency of using the package.
-In some circumstances, you can partially cache the results of compilation to a file
-(the `*.ji` files that live in your `~/.julia/compiled` directory) to reduce the burden
+In some circumstances, you can cache the results of compilation to files
+(the `*.ji` and `*.so` files that live in your `~/.julia/compiled` directory) to reduce the burden
 when your package is used.
```
mkitti, line 35: The parenthetical here breaks up the sentence flow and makes the idea harder to understand, so I moved this to the next sentence. `~/.julia/compiled` is the default location on most systems, but generally we need to refer to the Julia depot.

I then added some text to explain the trade off between caching more code and taking time to load it.

```suggestion
In some circumstances, you can cache the results of compilation to files to reduce the latency when your package is used. These files are the the `*.ji` and `*.so` files that live in the `compiled` directory of the Julia depot, usually located at `~/.julia/compiled`. However, if these files become large, loading them can be another source for latency. Julia needs time both to load and validate the cached compiled code. Minimizing the latency of using a package involves focusing on caching the compilation of code that is both commonly used and takes time to compile.
```

JamesWrigley: Added in 503139b (manually, so I could insert line breaks :) ).

timholy:
Thanks!

PR #378, opened 2024-04-04
Maybe `MaybeUndef` is Undef

chriselrod:
E.g., in Julia versions including https://github.com/JuliaLang/julia/pull/50257


mkitti:
This was addressed by https://github.com/timholy/SnoopCompile.jl/pull/379

PR #379, opened 2024-04-10
lazy loading of the JET integration

aviatesk:
JET sometimes ends up being incompatible with the latest version of Julia, and it can also prevent SnoopCompile.jl from loading. To get around this, this commit makes the JET integration lazy-loaded, so that it does not prevent SnoopCompile.jl from loading.


mkitti:
Could you fix the tests?

aviatesk:
I didn't dig into the failure on the stable version. We should investigate it, but let's do that in a separate PR.

timholy:
Thanks for doing this! FYI, given infinite time I'd be slimming down the number of Julia versions supported and the overall functionality of this package to probably just `@snoopr`, `@snoopi_deep`, and `@snoopl`. In other words, I'd support any efforts that aim in that direction, given suitable semver caretaking which you know all about.

timholy:
https://github.com/timholy/SnoopCompile.jl/issues/382#issuecomment-2059854034 pointed out that perhaps we should still have a `[compat]` entry for JET?

Issue #26, opened 2019-05-11
UndefVarError from getfield(Main, Symbol("##xx#xx")) symbols not defined


baggepinnen:
(On julia v1.2) I'm trying snoopcompile out for the first time. When snooping while running the tests of my package, the generated precompile file contains a lot of code with `getfield(Main, Symbol("##xx#xx"))`, like this
```julia
precompile(Tuple{typeof(MonteCarloMeasurements.gradient), getfield(Main, Symbol("##28#95")), MonteCarloMeasurements.Particles{Float64, 500}})
```
which causes 
```
julia> using MonteCarloMeasurements
[ Info: Recompiling stale cache file /local/home/fredrikb/.julia/compiled/v1.2/MonteCarloMeasurements/tHYBD.ji for MonteCarloMeasurements [0987c9cc-fe09-11e8-30f0-b96dd679fdca]
ERROR: Error while loading expression starting at /local/home/fredrikb/.julia/dev/MonteCarloMeasurements/src/MonteCarloMeasurements.jl:89
caused by [exception 1]
UndefVarError: ##28#95 not defined
Stacktrace:
 [1] _precompile_() at /local/home/fredrikb/.julia/dev/MonteCarloMeasurements/src/precompile_MonteCarloMeasurements.jl:11
```
I guess that the line in `runtests.jl` that lead to this is somewhere [here](https://github.com/baggepinnen/MonteCarloMeasurements.jl/blob/master/test/runtests.jl#L192)
Am I doing something wrong?

For reference, the code used to generate the precompile file is below
```julia
using SnoopCompile
SnoopCompile.@snoopc "/tmp/robust_opt_compiles.log" begin
    using MonteCarloMeasurements, Pkg
    include(joinpath(dirname(dirname(pathof(MonteCarloMeasurements))), "test", "runtests.jl"))
end
data = SnoopCompile.read("/tmp/robust_opt_compiles.log")
pc = SnoopCompile.parcel(reverse!(data[2]))
SnoopCompile.write("/tmp/precompile", pc)
```

timholy:
[...]

PR #380, opened 2024-04-14
Fix snoopc test

mkitti:
With Julia 1.6, a precompile statement is generated:

```julia
julia> using SnoopCompile

julia> logfile = joinpath(tempdir(), "anon.log")
"/tmp/anon.log"

julia> @snoopc logfile begin
           map(x->x^2, [1,2,3])
       end
Launching new julia process to run commands...
done.

julia> data = SnoopCompile.read(logfile)
(UInt64[0x00000000003d57dc, 0x00000000003d590b, 0x00000000016f8738], ["Tuple{Type{Base.Generator{I, F} where F where I}, Main.var\"#1#2\", Array{Int64, 1}}", "Tuple{typeof(Base.map), Function, Array{Int64, 1}}", "Tuple{typeof(Base.collect_similar), Array{Int64, 1}, Base.Generator{Array{Int64, 1}, Main.var\"#1#2\"}}"])

julia> read(logfile, String) |> println
4020491	"Tuple{typeof(Base.map), Function, Array{Int64, 1}}"
4020188	"Tuple{Type{Base.Generator{I, F} where F where I}, Main.var"#1#2", Array{Int64, 1}}"
24086328	"Tuple{typeof(Base.collect_similar), Array{Int64, 1}, Base.Generator{Array{Int64, 1}, Main.var"#1#2"}}"


julia> pc = SnoopCompile.parcel(reverse!(data[2]))
Dict{Symbol, Vector{String}} with 1 entry:
  :Base => ["precompile(Tuple{typeof(Base.map), typeof(identity), Array{Int64, 1}})"]
```

With Julia 1.10, the new precompile statement is not generated:

```julia
julia> using SnoopCompile

julia> logfile = joinpath(tempdir(), "anon.log")
"/tmp/anon.log"

julia> @snoopc logfile begin
           map(x->x^2, [1,2,3])
       end
Launching new julia process to run commands...
done.

julia> data = SnoopCompile.read(logfile)
(UInt64[0x00000000003bc942, 0x00000000013bae2e], ["Tuple{Type{Base.Generator{I, F} where F where I}, Main.var\"#1#2\", Array{Int64, 1}}", "Tuple{typeof(Base.collect_similar), Array{Int64, 1}, Base.Generator{Array{Int64, 1}, Main.var\"#1#2\"}}"])

julia> read(logfile, String) |> println
3918146	"Tuple{Type{Base.Generator{I, F} where F where I}, Main.var"#1#2", Array{Int64, 1}}"
[...]

PR #381, opened 2024-04-14
Make Cthulhu.jl a weakdep, implement CthulhuExt.jl

mkitti:
This is a continuation of https://github.com/timholy/SnoopCompile.jl/pull/376 for Cthulhu.jl after https://github.com/timholy/SnoopCompile.jl/pull/379

### DIFF
--- a/docs/src/jet.md
+++ b/docs/src/jet.md
@@ -137,5 +137,5 @@ Because SnoopCompile collected the runtime-dispatched `sum` call, we can pass it
 `report_callees` filters those calls which generate JET reports, allowing you to focus on potential errors.
 
 !!! note
-    JET integration is enabled only if JET.jl has been loaded into your main session.
-    This is why there's the `using JET` statement included in the example given.
+    JET integration is enabled only if JET.jl _and_ Cthulhu.jl have been loaded into your main session.
+    This is why there's the `using JET, Cthulhu` statement included in the example given.
--- a/docs/src/jet.md
+++ b/docs/src/jet.md
@@ -91,7 +91,7 @@ As always, you need to do the data collection in a fresh session where the calls
 After restarting Julia, we can do this:
 
 ```julia
-julia> using SnoopCompile
+julia> using SnoopCompile, JET, Cthulhu
 
 julia> using JET # this is necessary to enable the integration
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -30,7 +30,11 @@ logfile = joinpath(tempdir(), "anon.log")
 end
 data = SnoopCompile.read(logfile)
 pc = SnoopCompile.parcel(reverse!(data[2]))
-@test length(pc[:Base]) <= 1
+if Base.VERSION < v"1.10"
+    @test length(pc[:Base]) <= 1
+else
+    @test !haskey(pc, :Base)
+end
 
 # issue #29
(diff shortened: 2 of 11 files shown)


mkitti:
@aviatesk, could you please look this over?

docs/src/jet.md
```diff
@@ -91,7 +91,7 @@ As always, you need to do the data collection in a fresh session where the calls
 After restarting Julia, we can do this:
 
 ```julia
-julia> using SnoopCompile
+julia> using SnoopCompile, JET, Cthulhu
```
aviatesk, line 94: Please update the note at L139-L141 too.

test/runtests.jl
```diff
@@ -30,7 +30,11 @@ logfile = joinpath(tempdir(), "anon.log")
 end
 data = SnoopCompile.read(logfile)
 pc = SnoopCompile.parcel(reverse!(data[2]))
-@test length(pc[:Base]) <= 1
+if Base.VERSION < v"1.10"
+    @test length(pc[:Base]) <= 1
+else
+    @test !haskey(pc, :Base)
+end
```
aviatesk, line 37: Could you explain why this test case stopped working with v1.10 and later? Otherwise could you mark it as `broken` or simply leave it unchanged?

aviatesk (commented):
Thanks! LGTM albeit some comments below.

test/runtests.jl
mkitti: I provide some more detail on this in #380. This test was introduced by #27 to fix #26. In particular, #27 avoided making precompile statements for lines that contain anonymous functions. 

On Julia 1.6.7, three statements are generator from the `@snoopc` usage. Two of them contain anonymous functions and are correctly ignored.
```
4020491	"Tuple{typeof(Base.map), Function, Array{Int64, 1}}"
4020188	"Tuple{Type{Base.Generator{I, F} where F where I}, Main.var"#1#2", Array{Int64, 1}}"
24086328	"Tuple{typeof(Base.collect_similar), Array{Int64, 1}, Base.Generator{Array{Int64, 1}, Main.var"#1#2"}}"
```

On Julia 1.10.2, only two statements are generated and they are both correctly ignored.
```
3918146	"Tuple{Type{Base.Generator{I, F} where F where I}, Main.var"#1#2", Array{Int64, 1}}"
20688430	"Tuple{typeof(Base.collect_similar), Array{Int64, 1}, Base.Generator{Array{Int64, 1}, Main.var"#1#2"}}"
```

The only difference is that `Tuple{typeof(Base.map), Function, Array{Int64, 1}}` is already precompiled in Julia 1.10.2. This has nothing to do with the purpose of the test.

aviatesk: Alright, thanks so much for your explanation. All that sounds reasonable.

aviatesk (commented):
snoopi_deep_analysis.md and snoopr.md contain documentation on using `Cthulhu.ascend` for analysis. However with this PR `ascend` will not be available when loading `SnoopCompile` only. Could you update these documents to reflect this change? We also need to remove the `export` of `ascend` too.

mkitti:
I fixed the documentation and `ascend` export in [201f0c5](https://github.com/timholy/SnoopCompile.jl/pull/381/commits/201f0c57b51aefcf6361d3965ada3ea6dc146c3d).

timholy:
Awesome, thanks!

timholy:
xref #382

Issue #382, opened 2024-04-16
Next breaking release


timholy:
If I'm not mistaken, #379 and #381 are breaking changes. Before the next release, I'd propose we:
- [x] restrict to Julia 1.10 and higher (#384)
- [x] eliminate `SnoopCompileBot` (#383)
- [x] eliminate `SnoopPrecompile` (#384)
- [x] eliminate `@snoopc` and `@snoopi` (#384)
- [x] possibly rename `@snoopi_deep` to just `@snoopi` or `@snoop_infer` (#386)
- [x] update documentation (#389)

Other thoughts?

aviatesk:
Thanks for putting together the list. Let's focus on finishing it for now. We might come across additional tasks as we go along.

mkitti:
In https://github.com/timholy/SnoopCompile.jl/pull/379 we eliminated the combat declaration for JET.jl. I'm not sure if that was necessary since you can also keep a compact section as long as you also add the package to extras:
https://pkgdocs.julialang.org/v1/creating-packages/#Weak-dependencies

Also, since we are restricting to Julia 1.10 and above here though, perhaps we can restrict the JET.jl compatibility to tighter bounds that make sense for Julia 1.10?

mkitti:
Regarding release time, releasing this when Julia 1.11 comes out should be an objective. I'm not sure how long we have until then or if we have sufficient time to accomplish our goals above. While I see that we are trying to aggregate breaking changes and this is a good goal, I think that should be secondary to have a working package for a released and stable version of Julia.

timholy:
If needed I think we could get this out by next week. Given the number of [open milestones for Julia 1.11](https://github.com/JuliaLang/julia/issues?q=is%3Aopen+is%3Aissue+milestone%3A1.11) I don't think it will be hard to have this ready in time for 1.11.

aviatesk:
I support releasing version 3.0 as soon as we're ready, regardless of when 1.11 is released.

timholy:
The only really big job now is rewriting the documentation. SnoopCompile 2.x was written when we didn't cache native code, and I suspect that this assumption is pervasive in the documentation.

oscardssmith:
What's the status on this? I was just updating packages and realized that it is not currently possible to have latest versions of both JET and SnoopCompile installed.

timholy:
[...]

PR #384, opened 2024-04-16
Remove cruft

timholy:
This certainly isn't complete (e.g., the documentation will be updated separately), but it's a big chunk of #382:

- officially bump the version to 3.0.0
- require a minimum of Julia 1.10
- eliminate SnoopPrecompile
- delete a lot of version-specific cruft
- add two new extensions
- remove dependency on Requires


timholy:
I don't think this needs a detailed review, but I'd appreciate whether the general plan meets approval @aviatesk @mkitti

NEWS.md
```diff
@@ -30,3 +30,11 @@ This version implements major changes in how `parcel` works on the output of `@s
 
 Other changes:
 - A convenience utility, `timesum`, was introduced (credit to aminya)
+
+## Version 3
+
+Version 3 is a greatly slimmed repository, focusing just on those tools that are relevant for modern Julia development.
+Users of tools that were removed should stick with SnoopCompile 2.x.
+
+- The old `@snoopc` and `@snoopi` have been deleted.
```
mkitti, line 39: Could you add a note here about what they used to do and why they have been deleted?

NEWS.md
```diff
@@ -30,3 +30,11 @@ This version implements major changes in how `parcel` works on the output of `@s
...
+- SnoopCompileBot and SnoopPrecompile have been deleted.
```
mkitti, line 40: I would add a note here that SnoopPrecompile is now PrecompileTools.

SnoopPrecompile/README.md
```diff
@@ -1,5 +0,0 @@
-# SnoopPrecompile
```
mkitti, line 1: Maybe we could save this README to mainly link to PrecompileTools.jl?

SnoopCompileCore/src/snoopi_deep.jl
```diff
@@ -1,3 +1,5 @@
+export @snoopi_deep
```
mkitti, line 1: Were you going to alias `@snoopi_deep` as `@snoopi`?

mkitti (approved):
I would leave some additional breadcrumbs about the changes and the removed features.

aviatesk (approved):
Glad to see old code removed! I think we probably want to rename `@snoopi_deep` to `@snoopi`, while also deprecating `@snoopi_deep` and making it an alias for the new `@snoopi`?

timholy:
How about a more wholesale renaming?
- `@snoopi_deep` -> `@snoop_infer`
- `@snoopl` -> `@snoop_llvm`
- `@snoopr` -> `@snoop_inval`

aviatesk:
> How about a more wholesale renaming?
> 
> * `@snoopi_deep` -> `@snoop_infer`
> * `@snoopl` -> `@snoop_llvm`
> * `@snoopr` -> `@snoop_inval`

That sounds good to me. I'm all for that.

SnoopCompileCore/src/snoopi_deep.jl
timholy: In a separate PR, yes. Let's keep this one focused on deleting stuff.

timholy:
Further changes, like renaming the main data-collection macros, are things I intend for a future PR.

PR #386, opened 2024-07-02
Rename the macros

timholy:
The macro names are now longer but clearer:
- `@snoop_invalidations`
- `@snoop_inference`
- `@snoop_llvm`

Thoughts?

Because I also renamed the files and git does not gracefully track file renaming, this seems like a much larger change than it is. EDIT: github, however, seems to track these OK.


timholy:
Will require https://github.com/JuliaPy/PyPlot.jl/pull/589 to pass doctests, but let's merge anyway.

PR #389, opened 2024-07-07
Rewrite the docs

timholy:
This aims to rewrite the documentation loosely in the style advocated by [divio](https://docs.divio.com/documentation-system/).
It acknowledges the existence of PrecompileTools and focuses on how you use SnoopCompile to understand why precompilation isn't working the way you'd like.


timholy (commented):
At long last (*looong* last), this is almost ready. I need to work through one of the demos ("Using `@snoop_inference` to improve inferability") and likely fix some issues at the end, but otherwise I think this is ready for anyone who wants to review it.

timholy:
You can click on the `documenter/deploy` link above or see https://timholy.github.io/SnoopCompile.jl/previews/PR389/

timholy:
Hmm, Documenter seems like it might have trouble until we register this 😢

timholy:
OK, I think this is finally ready. I anticipate merging and releasing SnoopCompile 3.0.0 later today.

PR #394, opened 2024-09-11
Docs: add `maxdepth=100` arg to print_tree

IanButterworth:
Without this it's common to get truncation because the default is 5.

And discoverability isn't great because `AbstractTrees.print_tree` isn't documented https://juliacollections.github.io/AbstractTrees.jl/stable/search/?q=print_tree
but has a docstring that shows this arg https://github.com/JuliaCollections/AbstractTrees.jl/blob/af3fabd36631efe361bc8e1ba45f9b84e898b459/src/printing.jl#L17


timholy:
Thanks!

PR #398, opened 2024-10-13
Fix SCPyPlotExt

Sbozzolo:
The `SCPyPlotExt` had three problems:

- the default argument to `pdata` is `Profile.fetch()`, but `Profile` was not imported
- the extension was not adding methods to `SnoopCompile.pgdsgui`
- `prep_ri` was failing with `UndefVarError: lookup_firstip! not defined`

As a result, the tutorial on `pgdsgui` was broken.

This commit fixes these issues. In the process, I also fixed syntax highlighing in a couple of places in the documentation.

### DIFF
--- a/ext/SCPyPlotExt.jl
+++ b/ext/SCPyPlotExt.jl
@@ -1,5 +1,7 @@
 module SCPyPlotExt
 
+import Profile
+
 using SnoopCompile
 using SnoopCompile: MethodLoc, InferenceTimingNode, PGDSData, lookups
 using PyPlot: PyPlot, plt, PyCall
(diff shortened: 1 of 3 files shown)


ext/SCPyPlotExt.jl
```diff
@@ -48,17 +50,17 @@ function pgdsgui(ax::PyCall.PyObject, ridata::AbstractVector{Pair{Union{Method,M
...
-function pgdsgui(args...; kwargs...)
+function SnoopCompile.pgdsgui(args...; kwargs...)
     fig, ax = plt.subplots()
     pgdsgui(ax, args...; kwargs...), ax
 end
 
 function prep_ri(tinf::InferenceTimingNode, pdata=Profile.fetch(); lidict=lookups, consts, by, kwargs...)
-    lookup_firstip!(lookups, pdata)
+    SnoopCompile.lookup_firstip!(lookups, pdata)
     return runtime_inferencetime(tinf, pdata; lidict, consts, by)
```
charleskawczynski, line 64:
```suggestion
    SnoopCompile.lookup_firstip!(lookups, pdata)
    return SnoopCompile.runtime_inferencetime(tinf, pdata; lidict, consts, by)
```
I don't see an `export` for `runtime_inferencetime` either

timholy:
Many thanks!

PR #413, opened 2025-03-20
Fix test failures on 1.11

timholy:


timholy:
I now have a much clearer understanding of *what* is going wrong, but I still lack an understanding of *why*.

In case later changes fix it, the report in question is https://github.com/timholy/SnoopCompile.jl/actions/runs/13989097828/job/39168795257?pr=413

Here's the logic: https://github.com/timholy/SnoopCompile.jl/blob/teh/debug_1.11/test/testmodules/Stale/StaleC/src/StaleC.jl should "heal" invalidations in the `build_stale` call chain. So the appearance of `use_stale` in the [recompilation result](https://github.com/timholy/SnoopCompile.jl/blob/a8f0490237f0c26ac50d35baca38f1298f532b80/test/snoop_inference.jl#L873-L876)` below
```
InferenceTimingNode: 0.038472/0.054491 on Core.Compiler.Timings.ROOT() with 2 direct children
├─ InferenceTimingNode: 0.000088/0.000097 on StaleB.useA() with 1 direct children
│  └─ InferenceTimingNode: 0.000009/0.000009 on getproperty(StaleA::Module, stale::Symbol) with 0 direct children
└─ InferenceTimingNode: 0.015922/0.015922 on StaleA.use_stale(::Vector{Any}) with 0 direct children
unexpected result: 2 backedges
```
is surprising. "0 direct children" seems to suggest that nothing "needed" it, and a test a little higher up in the file https://github.com/timholy/SnoopCompile.jl/blob/a8f0490237f0c26ac50d35baca38f1298f532b80/test/snoop_inference.jl#L858 explicitly checks that there is a valid compiled instance. So I can't understand why this method is being recompiled (likely triggered by the https://github.com/timholy/SnoopCompile.jl/blob/teh/debug_1.11/test/snoop_inference.jl#L875 line).
Locally, this happens only with `coverage=true`, with `coverage=false` it behaves as I expect.

timholy:
Summary of a discussion on slack: this is a consequence of avoiding cached native code:

- https://github.com/JuliaLang/julia/pull/52123
- https://github.com/JuliaLang/julia/pull/53439
- https://github.com/JuliaLang/julia/pull/53457

in conjunction with the more recent (Julia 1.11.3 and Julia 1.11.4) removal of the inferred code:

https://github.com/JuliaLang/julia/pull/56749

The plan is to revert the latter on Julia 1.11, hopefully landing in Julia 1.11.5. Meanwhile, I'll edit the test to compensate.

timholy:
Ah, lovely.

PR #416, opened 2025-03-23
Add docs for `isprecompilable`

### LINKED ISSUE #411: Add user-documentation for `isprecompilable`
Over at Clima we're running into some latency issues, and a lot of our code seems to be, according to the docs, not "naively precompilable". We generated flame graphs using `@snoop_inference`, but most of them are [red](https://github.com/CliMA/ClimaCore.jl/pull/2225#issuecomment-2730444846). There isn't really any mention of what conditions are needed to make code precompilable.

Our code satisfies a lot of other important properties:
 - it's type stable as I can see (we pass JET tests)
 - not allocating
 - No type piracy
 - we've reduced loads of method invalidations

Any tips are greatly appreciated.

We have experiments that, as far as I can tell, result in the compiler allocating ~100 GB of memory and takes ~7 minutes to compile for the first time (99% of the time is compilation). I can't easily make a "simple reproducer" for this, but we are definitely experiencing development pains as this is difficult to develop with.

timholy:
This also enhances its implementation to make it more flexible.

Closes #411

### DIFF
--- a/src/parcel_snoop_inference.jl
+++ b/src/parcel_snoop_inference.jl
@@ -306,7 +306,7 @@ Main.B
 Now let's run these methods to generate some compiled `MethodInstance`s:
 
 ```jldoctest isprecompilable
-julia> A.a(3.2)          # Float64 is not "owned" by A, but A loads Base so knows about it
+julia> A.a(3.2)          # Float64 is not "owned" by A, but A loads Base so A knows about it
 3.2
 
 julia> A.a(B.BType())    # B.BType is not known to A
--- a/src/parcel_snoop_inference.jl
+++ b/src/parcel_snoop_inference.jl
@@ -276,6 +276,10 @@ If `mi` has been compiled by the time its defining module "closes" (the final
 Julia will automatically include this specialization in that module's precompile
 cache.
 
+!!! tip
+    If `mi` is a MethodInstance corresponding to `f(::T)`, then calling `f(x::T)`
+    before the end of the module definition suffices to force compilation of `mi`.
+
 If you'd like to cache it but `isprecompilable(mi)` returns `false`, you need to
 identify a module `mod` for which `isprecompilable(mod, mi)` returns `true`.
 However, just ensuring that `mi` gets compiled within `mod` may not be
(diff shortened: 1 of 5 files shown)


src/parcel_snoop_inference.jl
```diff
@@ -260,26 +260,103 @@ end
...
+
+Determine whether `mi` is able to be precompiled within `mod`. This requires
+that all the types in `mi`'s specialization signature are "known" to `mod`. See
+[`SnoopCompile.known_type`](@ref) for more information.
+
+`isprecompilable(mi)` sets `mod` to the module in which the corresponding method
+was defined. If `mod ∈ excluded_modules`, then `isprecompilable` returns `false`.
+
+If `mi` has been compiled by the time its defining module "closes" (the final
+`end` of the module definition) and `isprecompilable(mi)` returns `true`, then
```
charleskawczynski, line 275: > If `mi` has been compiled by the time its defining module "closes" (the final
`end` of the module definition)

When might this happen? Would it mean that the defining module would need to call `mi`?

src/parcel_snoop_inference.jl
```diff
@@ -260,26 +260,103 @@ end
...
+       struct BType end    # this type is not known to A
+       b(x) = x
+       end
+Main.B
+```
+
+Now let's run these methods to generate some compiled `MethodInstance`s:
+
+```jldoctest isprecompilable
+julia> A.a(3.2)          # Float64 is not "owned" by A, but A loads Base so knows about it
```
charleskawczynski, line 309:
```suggestion
julia> A.a(3.2)          # Float64 is not "owned" by A, but A loads Base so A knows about it
```

charleskawczynski (approved):
I posted a few comments / suggestions, but overall looks good, thank you!

src/parcel_snoop_inference.jl
timholy: If `mi` corresponds to `f(x)`, then just calling `f(x)` before the module closes will do it. For the inferrable callees of this method, it won't matter if it's inside or outside a `@compile_workload` block. For any runtime-dispatched ones, it will matter a lot. But if you're thinking about a single "flame" on the flamegraph, that's irrelevant, because one flame = the inferrable callees of the entrypoint (base of the flame).

PR #419, opened 2025-04-15
Update `@snoop_inference` for Julia 1.12

timholy:
This leverages the [new timing fields of `CodeInstance`](https://github.com/JuliaLang/julia/pull/57074) to replace the old `Core.Compiler.Timings` module. There is a bit of loss particularly in the domain of constant-propagation, which is illustrated in https://gist.github.com/timholy/9a64b27c1932bb414e69b8fe48284b5e. (Read `testcase.jl` before looking at the results.) As a consequence there are a small handful of failing tests locally, but the vast majority pass.

This requires https://github.com/JuliaLang/julia/pull/58124 or alternatively (with some modification to this PR) https://github.com/JuliaLang/julia/pull/58123.

### DIFF
--- a/SnoopCompileCore/src/snoop_inference.jl
+++ b/SnoopCompileCore/src/snoop_inference.jl
@@ -2,24 +2,23 @@ export @snoop_inference
 
 const snoop_inference_lock = ReentrantLock()
 const newly_inferred = CodeInstance[]
+const inference_entrance_backtraces = []
 
 function start_tracking()
     iszero(snoop_inference_lock.reentrancy_cnt) || throw(ConcurrencyViolationError("already tracking inference (cannot nest `@snoop_inference` blocks)"))
     lock(snoop_inference_lock)
     empty!(newly_inferred)
+    empty!(inference_entrance_backtraces)
     ccall(:jl_set_newly_inferred, Cvoid, (Any,), newly_inferred)
-    # return ccall(:jl_log_dispatch_backtrace, Any, (Cint,), 1)
-    empty!(Base.dispatch_backtrace)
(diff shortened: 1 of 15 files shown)


SnoopCompileCore/src/snoop_inference.jl
```diff
@@ -1,100 +1,26 @@
...
-    ttree = timingtree(t)
-    it, start_time, ttree_children = ttree::Tuple{InferenceTiming, Float64, Vector{Any}}
-    root = InferenceTimingNode(it, start_time, t.bt)
-    addchildren!(root, t, ttree_children)
-    return root
+const snoop_inference_lock = ReentrantLock()
+const newly_inferred = CodeInstance[]
+
+function start_tracking()
+    islocked(snoop_inference_lock) && error("already tracking inference (cannot nest `@snoop_inference` blocks)")
```
vtjnash, line 7: (this really should be a Base feature, since implementing it indirectly in packages is not very robust)
```suggestion
    snoop_inference_lock.reentrancy_cnt == 1 || throw(ConcurrencyViolationError("already tracking inference (cannot nest `@snoop_inference` blocks)"))
```

SnoopCompileCore/src/snoop_inference.jl
```diff
@@ -1,100 +1,26 @@
...
-        start_deep_timing()
-        try
-            $(esc(cmd))
-        finally
-            stop_deep_timing()
-        end
-        finish_snoop_inference()
-    end
+function stop_tracking()
+    @assert islocked(snoop_inference_lock)
```
vtjnash, line 18: nit: more concisely
```suggestion
    Base.assert_havelock(snoop_inference_lock)
```

PR #421, opened 2025-07-25
Handle untracked causes of invalidation

timholy:
There seem to be rare cases where edge invalidations tagged with
`"verify_method"` do not list their cause. This adds a fallback target
to store such invalidations.

Also fixes an issue with the ordering of insertions that triggered an
assert. CC @LilithHafner


timholy:
Test failure sources are from #419 not yet being merged, and from the JET integration test.

PR #430, opened 2025-09-19
Support `Jet@0.10`

oscardssmith:
Tests pass locally at least


timholy:
https://github.com/JuliaRegistries/General/pull/138888

PR #445, opened 2025-10-17
Fix potential inifinite loop in `add_children!()`

### LINKED ISSUE #444: Inifinite loop in add_children!()
I have a case where I get stuck in an infinite loop here
https://github.com/JuliaDebug/SnoopCompile.jl/blob/a0a03f3b2943a6275cd58aa3dec64d172bda264e/SnoopCompileCore/src/snoop_inference.jl#L127-L155

I think what's happening is that `be ∉ handled` is true the first time it is tested, for `k == j`. That means that `j` never increments, and the loop never ends. I don't understand what's being done here, but if I change line 137 to
```julia
k ≤ j && continue   # don't get caught in cycles
```
instead of `k < j`, then there's no infinite loop, and the output looks reasonable. Is this a correct fix?

johnomotani:
A method can call itself, so need to skip `k == j` as well as `k < j`.

Fixes #444.

### DIFF
--- a/SnoopCompileCore/src/snoop_inference.jl
+++ b/SnoopCompileCore/src/snoop_inference.jl
@@ -134,7 +134,7 @@ function addchildren!(parent::InferenceTimingNode, cis, backedges, miidx, backtr
         while true
             found = false
             for k in backedges[j]
-                k < j && continue   # don't get caught in cycles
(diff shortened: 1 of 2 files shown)


timholy:
Thanks for your contribution!

timholy:
https://github.com/JuliaRegistries/General/pull/142159

PR #460, opened 2026-02-20
Handle `nothing` methods in SCPrettyTablesExt

### LINKED ISSUE #449: `nothing` in `report_invalidations`
The [julia-invalidations](https://github.com/julia-actions/julia-invalidations) script currently crashes.

```julia
ERROR: LoadError: FieldError: type Nothing has no field `name`; Nothing has no fields at all.
Stacktrace:
  [1] getproperty
    @ ./Base_compiler.jl:54 [inlined]
  [2] #12
    @ ~/.julia/packages/SnoopCompile/sSjKE/ext/SCPrettyTablesExt.jl:35 [inlined]
```

I think the problem is the same as in #447: `inv.method` below can be `nothing`.

https://github.com/JuliaDebug/SnoopCompile.jl/blob/6cf031f62ba8b89968adbf39650c2ba395ddb085/ext/SCPrettyTablesExt.jl#L34-L39

lgoettgens:
This resolves https://github.com/JuliaDebug/SnoopCompile.jl/issues/449 by not trying to access fields of `nothing`, instead showing a fixed string.

Of course, it would be even better to understand where these `nothing` entries come from, but I think doing this obvious fix already helps quite a lot of people.

### DIFF
--- a/ext/SCPrettyTablesExt.jl
+++ b/ext/SCPrettyTablesExt.jl
@@ -32,10 +32,10 @@ function SnoopCompile.report_invalidations(io::IO = stdout;
         Float16(100 * inv / sum_invs)
     end
     meth_name = map(trees) do inv
-        "$(inv.method.name)"
+        isnothing(inv.method) ? "unknown" : "$(inv.method.name)"
     end
     fileinfo = map(trees) do inv
-        "$(process_filename(string(inv.method.file))):$(inv.method.line)"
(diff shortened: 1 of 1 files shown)


lgoettgens:
Bump @aviatesk @timholy

lgoettgens:
> Bump @aviatesk @timholy

Gentle bump

aviatesk (approved):

schillic:
@aviatesk: Thanks for merging. Could you please also make a new release so this fix gets published?

aviatesk:
v3.2.6 is on the way.

